PR #50, opened 2015-08-28
Add PingHandler back to pipeline.

willsalz:
- Also fix one TCHannel service getter to match style guide, which broke `tcurl`
- Fix KeyValue thrift file comments
- run code lint, which removed an unnecessary cast from the
  ChannelManager


willsalz:
cc @truncs @brandoniles

truncs:
travis CI fails when running the thrift command. are you sure travis is able to install thrift-compiler ?

willsalz:
I have been unsuccessful thus far in installing `thrift`. I’ve actually been working with somebody from travis-ci and we haven’t root caused it. It *should* work in a containerized environment but that fails. I tried in a non-container environment as well, which supposedly should work, but Ubuntu 12.0.4 doesn’t have a `thrift-compler` package. I’m still digging into it. One last solution would be to build a custom docker image and use their dockerized container.

truncs:
There are backports of the packge by other people if that is possible 

http://people.apache.org/~jfarrell/thrift/0.9.1/contrib/deb/ubuntu/12.04/

truncs:
@willsalz lets merge this and track the travis/thrift in a different issue

PR #52, opened 2015-08-29
Expose TTL and Response Code

willsalz:
Why
====
We weren't exposing Request TTLs nor Response Error code.

Misc
====
- Move required fields for Request and Response to the constructor
- Definitely needs more tests…

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
@@ -24,6 +24,7 @@ package com.uber.tchannel.api;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 public final class Request<T> {
 
@@ -82,10 +83,13 @@ public final class Request<T> {
 
     public static class Builder<U> {
 
-        private static final long DEFAULT_TTL = 100;
-
         private String service;
-        private long ttl = DEFAULT_TTL;
+        /*
+        * Time To Live in milliseconds. Defaults to a reasonable 100ms, as requests *cannot* be made without a TTL set.
+        *
+        * see: https://github.com/uber/tchannel/blob/master/docs/protocol.md#ttl4
+        */
+        private long ttl = 100;
         private Map<String, String> transportHeaders = new HashMap<>();
         private String endpoint;
         private Map<String, String> headers = new HashMap<>();
@@ -97,8 +101,22 @@ public final class Request<T> {
             this.endpoint = endpoint;
         }
 
+        /**
+         * @param ttl TTL in milliseconds
+         * @return Builder
+         */
         public Builder<U> setTTL(long ttl) {
             this.ttl = ttl;
+            return this;
+        }
+
+        /**
+         * @param ttl      TTL in `timeUnit` units
+         * @param timeUnit time unit for the `ttl`
+         * @return Builder
+         */
+        public Builder<U> setTTL(long ttl, TimeUnit timeUnit) {
+            this.ttl = TimeUnit.MILLISECONDS.convert(ttl, timeUnit);
             return this;
         }
 
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
@@ -28,6 +28,7 @@ import java.util.Map;
 public final class Request<T> {
 
     private final String service;
+    private final long ttl;
     private final Map<String, String> transportHeaders;
     private final String endpoint;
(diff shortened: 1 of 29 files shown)


tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
```diff
@@ -76,18 +82,23 @@ public String toString() {
 
     public static class Builder<U> {
 
+        private static final long DEFAULT_TTL = 100;
```
truncs: nitpick: this contant makes little sense, you can just have the default value of var ttl itself as 100.

tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
```diff
@@ -76,18 +82,23 @@ public String toString() {
...
+
         private String service;
+        private long ttl = DEFAULT_TTL;
         private Map<String, String> transportHeaders = new HashMap<>();
         private String endpoint;
         private Map<String, String> headers = new HashMap<>();
         private U body;
 
-        public Builder(U body) {
+        public Builder(U body, String service, String endpoint) {
```
truncs: this is good change, since all these are required fields for a Request

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -241,6 +244,9 @@ protected void initChannel(SocketChannel ch) throws Exception {
                         ch.pipeline().addLast("InitRequestInitiator", new InitRequestInitiator());
                     }
 
+                    // Handle PingRequest
+                    ch.pipeline().addLast("PingHandler", new PingHandler());
```
truncs: this change is present in your other pull request #50 as well ?

tchannel-core/src/main/java/com/uber/tchannel/handlers/ResponseRouter.java
```diff
@@ -87,8 +88,10 @@ protected void messageReceived(ChannelHandlerContext ctx, RawResponse rawRespons
         ResponsePromise<?> responsePromise = this.messageMap.remove(rawResponse.getId());
```
truncs: I have a unrelated comment for ResponsePromise, does ResponsePromise fails for a errored Response call ?

truncs:
@willsalz few comments, rest looks good.

truncs:
Also these are breaking the API  cc @brandoniles

brandoniles:
Are there units for TTL? If it's standard enough to go unstated or it's defined by TChannel, it's fine as it is. I can't tell myself from looking, though.

willsalz:
TTL is [always in milliseconds](https://github.com/uber/tchannel/blob/master/docs/protocol.md#ttl4). One unfortunate ambiguity is that TTL and a single Request timeout may be different things.

tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
willsalz: yeah this was a good tip from Effecitive Java :smiley:

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
willsalz: Yeah it is actually; this is rebased off of that PR.

tchannel-core/src/main/java/com/uber/tchannel/handlers/ResponseRouter.java
willsalz: So, the answer is: 'it depends'. If the Response is an 'error' response then yes. In the face of a fatal protocol error: 'no'.

There are a few scenarios under which we will fail:
- no response back: we don't gracefully handle this scenario
- Protocol `ErrorMessage`: we don't gracefully handle this scenario
- Response with `Error` response code: we successfully resolve the Promise.

I'll document these error edge-cases and see what we can do to cleanup our failure cases.

brandoniles:
LGTM

tchannel-core/src/main/java/com/uber/tchannel/api/Request.java
willsalz: agreed; but I wanted to document the value somehow…I'll just set the TTL to a val and add a wordy-docstring.

tchannel-core/src/main/java/com/uber/tchannel/handlers/ResponseRouter.java
willsalz: filed #55

truncs:
@willsalz LGTM ! Feel free to merge, you might have to rebase to master and make some changes before you can merge.

PR #57, opened 2015-09-01
Prepare Project for Maven Central Upload

willsalz:
Why
====
We would like to provide `tchannel-java` as a package that is easily integrated into traditional Java Maven projects e.g. Hosted in a Central Repository.

What
====
Since one cannot simply upload a project to Maven Central, we need to go
through a 3rd party 'host'. There are a few available but it appears as
though both Netty and Google Guava use Sonatype so I went with them.
There are a few requirements, and the full guide lives here:
http://central.sonatype.org/pages/ossrh-guide.html.

Misc
------
- Update the artifactId to `tchannel-java`
- Move versiont to SemVer
- Update docs everywhere
- Change tchannel-benchmark name to `tchannel-benchmark` from the
  generic name from the JMH template

reviewers: @truncs @brandoniles @zsong03
cc: @sectioneight @jwolski

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,13 +56,13 @@ System.out.println(response);
 #### PingServer
 ```bash
 mvn package
-java -cp tchannel-example/target/tchannel-example-1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingServer -p 8888
+java -cp tchannel-example/target/tchannel-example-0.1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingServer -p 8888
 ```
 
 #### PingClient
 ```bash
 mvn package
-java -cp tchannel-example/target/tchannel-example-1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingClient -h localhost -p 8888 -n 1000
+java -cp tchannel-example/target/tchannel-example-0.1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingClient -h localhost -p 8888 -n 1000
 ```
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -56,13 +56,13 @@ System.out.println(response);
 #### PingServer
 ```bash
 mvn package
-java -cp tchannel-example/target/tchannel-example-0.1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingServer -p 8888
+java -cp tchannel-example/target/tchannel-example.jar com.uber.tchannel.ping.PingServer -p 8888
 ```
 
 #### PingClient
 ```bash
 mvn package
-java -cp tchannel-example/target/tchannel-example-0.1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingClient -h localhost -p 8888 -n 1000
+java -cp tchannel-example/target/tchannel-example.jar com.uber.tchannel.ping.PingClient -h localhost -p 8888 -n 1000
 ```
 
 ## Contributing
--- a/pom.xml
+++ b/pom.xml
@@ -26,11 +26,11 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.uber.tchannel</groupId>
-    <artifactId>tchannel-java</artifactId>
-    <version>0.1.0-SNAPSHOT</version>
+    <artifactId>tchannel</artifactId>
+    <version>0.1.1-SNAPSHOT</version>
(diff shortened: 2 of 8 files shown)


brandoniles:
This is beyond my maven knowledge, but as far as I can tell nothing looks out of place.

Is there a way to test the maven integration before going through the 3rd party?

pom.xml
```diff
@@ -26,9 +26,36 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>com.uber.tchannel</groupId>
-    <artifactId>tchannel</artifactId>
+    <artifactId>tchannel-java</artifactId>
```
truncs: why not just call it tchannel ?

pom.xml
```diff
@@ -26,9 +26,36 @@
...
+    <version>0.1.0-SNAPSHOT</version>
```
truncs: with this versioning scheme only 2 release are allowed per version, the first one the SNAPSHOT or developer release and the next is 0.1.0 which is the actual release.

pom.xml
```diff
@@ -26,9 +26,36 @@
...
+    <licenses>
+        <license>
+            <name>MIT License</name>
+            <url>http://www.opensource.org/licenses/mit-license.php</url>
+        </license>
+    </licenses>
+
+    <developers>
+        <developer>
+            <name>Will Salisbury</name>
```
truncs: can we have this as tchannel contributors (its hard to manage all developer contributions like this going forward) and instead of your personal email some general google group where people can actually ask questions ?

pom.xml
willsalz: yeah…probably should.

pom.xml
willsalz: so, my understanding is that we should/could continually update SNAPSHOT builds i.e. un-versioned nightlies, and then release with traditional SemVer semantics e.g. Major.Minor.Patch. 

What do you think / propose as an alternative?

truncs: Its fine to update SNAPSHOT as long as people don't add it as the depedency, that is why a lot of people do multiple Alpha and Beta releases before releasing a final version.

pom.xml
willsalz: yep. I'll reach out to the OSS folks about a mailing list…I think they're still trying to figure that out in the context of the larger TChannel / Hyperbahn effort.

pom.xml
junchaowu: SNAPSHOT usually is version before 1.0. But that's just convention. If java tchannel isn't ready or stable yet, we can still have SNAPSHOT.

pom.xml
willsalz: Emailed. Some of them are also CC'd…

README.md
```diff
@@ -56,13 +56,13 @@ System.out.println(response);
 #### PingServer
 ```bash
 mvn package
-java -cp tchannel-example/target/tchannel-example-1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingServer -p 8888
+java -cp tchannel-example/target/tchannel-example-0.1.0-SNAPSHOT.jar com.uber.tchannel.ping.PingServer -p 8888
```
truncs: this should be 0.1.1 or better just someway generalize it.

pom.xml
```diff
@@ -27,8 +27,35 @@
 
     <groupId>com.uber.tchannel</groupId>
     <artifactId>tchannel</artifactId>
+    <version>0.1.1-SNAPSHOT</version>
```
truncs: wait so did we already do a 0.1.0 release ?

README.md
willsalz: Are you talking bout the built projects jars? I can setup a maven shade execution to just create a jar for each of our examples (server and client) which will make running and documenting them easier

pom.xml
willsalz: kind of. we 'tagged a release' but since we weren't uploading / releasing anything to the Central Repository it was kind of a fake release.

truncs: okay sounds good.

README.md
truncs: feel free to merge this, this shouldn't be blocking this diff

pom.xml
willsalz: no email lists setup yet. OSS recommendation is to just use my email for now.

PR #59, opened 2015-09-06
add a parametrized ping pong benchmark

truncs:
The results look pretty bad with delays of 1ms and 10ms. I ran it on my macbook pro 8 core 16GB of RAM

```
Benchmark                          (sleepTime)   Mode  Cnt      Score      Error  Units
PingPongServerBenchmark.benchmark            0  thrpt   10  10240.926 ± 1091.874  ops/s
PingPongServerBenchmark.benchmark            1  thrpt   10    798.320 ±   82.839  ops/s
PingPongServerBenchmark.benchmark           10  thrpt   10     85.511 ±    2.623  ops/s
```

### DIFF
--- a/tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
+++ b/tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
@@ -32,6 +32,7 @@ import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
 import org.openjdk.jmh.annotations.Level;
 import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Param;
 import org.openjdk.jmh.annotations.Scope;
 import org.openjdk.jmh.annotations.Setup;
 import org.openjdk.jmh.annotations.State;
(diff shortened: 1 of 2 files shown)


tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
```diff
@@ -83,9 +89,6 @@ public void benchmark() throws Exception {
                 request,
                 Pong.class
         );
-
-        f.get();
```
willsalz: Why are you removing the get?

truncs: so that the benchmarking thread doesn't get blocked.

truncs: The strange part is even without f.get I can't get the performance to upto what I wanted. This suggest that there is a performance bottleneck in just making the calls.

Also while logging stuff on request router side, there were performance bottlenecks over there as well but this benchmark is not able to expose that part.

willsalz: I believe this is related to serial handling of requests in the RequestDispatcher…I'll file a bug and see if I can put in a quick fix w/ an executor / ThreadPool

PR #60, opened 2015-09-07
use reflections to determine the request type and response type

truncs:
cc @willsalz @brandoniles

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/util/ReflectionUtils.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/util/ReflectionUtils.java
@@ -1,0 +1,103 @@
+/*
+ * Copyright (c) 2015 Uber Technologies, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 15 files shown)


tchannel-core/src/main/java/com/uber/tchannel/util/ReflectionUtils.java
```diff
@@ -0,0 +1,103 @@
...
+        } else if (type instanceof GenericArrayType) {
+            Type componentType = ((GenericArrayType) type).getGenericComponentType();
+            Class<?> componentClass = getClass(componentType);
+            if (componentClass != null) {
+                return Array.newInstance(componentClass, 0).getClass();
+            } else {
+                return null;
+            }
+        } else {
+            return null;
```
brandoniles: What's the effect of returning null here? Will a client have a good idea what's going on if this happens?

truncs: it should never return a null if the POJOs are specified in RequestHandler which is enforced by our class signature.

willsalz:
LGTM. Although, There has got to a better way where we don't need to support this ourselves. I've checked out the Apache Commons - Lang package and might have what we need. I can play around ( later… ) and see if we can move this logic out of TChannel-Java

truncs:
@willsalz Gauva's TypeToken would do the same thing I suppose, https://github.com/google/guava/wiki/ReflectionExplained#typetoken

In any case this break the API since this as RequestHandler as abstract class, so it would be nice if we can merge this and figure out the implementation details later ?

brandoniles:
Instead of converting the interface to an abstract class, you can consider adding the abstract class below the interface as a skeletal implementation. This could allow for more flexibility in the API.

willsalz:
+1 to a `DefaultRequestHandler` Abstract implementation of the RequestHandler.

willsalz:
@truncs any new thoughts on @brandoniles and my comments r.e. a Default ABC + Keeping the Interface?

PR #69, opened 2015-09-15
Fixing Javadoc build failures and checkstyle inconsistencies.

willsalz:
We're preparing to release v0.1.2, our first release to the central
repositories, which means we're going through a lot of new build phases
e.g. javadocs, gpg signing, maven releaseing, etc. I'm learning a lot
and hopefully we'll nail-down a quasi-sane workflow. But until then,
it's a lot of trial-and-error followed by frantic commits.

cc @brandoniles @truncs

### DIFF
--- a/tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
+++ b/tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
@@ -22,8 +22,8 @@
 
 package com.uber.tchannel.benchmarks;
 
-import com.uber.tchannel.api.DefaultRequestHandler;
 import com.uber.tchannel.api.Request;
+import com.uber.tchannel.api.DefaultRequestHandler;
 import com.uber.tchannel.api.Response;
 import com.uber.tchannel.api.ResponseCode;
 import com.uber.tchannel.api.TChannel;
@@ -57,11 +57,7 @@ public class PingPongServerBenchmark {
     TChannel client;
     int port;
 
-    @Param({
-            "0",
-            "1",
-            "10"
-    })
+    @Param({ "0", "1", "10" })
     private int sleepTime;
 
     public static void main(String[] args) throws RunnerException {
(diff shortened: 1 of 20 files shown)


tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
```diff
@@ -57,7 +57,11 @@
     TChannel client;
     int port;
 
-    @Param({ "0", "1", "10" })
+    @Param({
```
truncs: what are these changes for ?

tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
```diff
@@ -108,6 +112,21 @@ public void teardown() throws Exception {
         this.channel.shutdown();
     }
 
+    @AuxCounters
```
truncs: is this a style violation ?

tchannel-core/src/main/java/com/uber/tchannel/api/DefaultRequestHandler.java
```diff
@@ -26,8 +26,10 @@
 
 public abstract class DefaultRequestHandler<T, U> implements RequestHandler<T, U> {
 
-    private TypeToken<T> requestType = new TypeToken<T>(getClass()) { };
-    private TypeToken<U> responseType = new TypeToken<U>(getClass()) { };
+    private TypeToken<T> requestType = new TypeToken<T>(getClass()) {
```
truncs: As long as we conform the automated style checks I don't think we should be making these changes. Its just not scalable, so either push these stylechecks to the automated ones or lets not fix them.

truncs:
@willsalz why don't we just stay on 0.1.2-SNAPSHOT ? The project still looks like its in alpha phase.

willsalz:
Yeah I can remove the auto-formatting from the diff.

And in terms of the version number: in order to publish to the central maven repository we have to push non-SNAPSHOT builds **unless** we have our clients put special declarations in their pom.xml stating specifically *which* SNAPSHOT repository to allow ( in this case, Sonatype's ). So we can keep a 'zero-major' SemVer version e.g. pre-1.0 and deploy those with standard machinery common to uploading projects to the repo.

truncs:
that is why I like the netty style versioning better where each release is tagged with either Alpha, Beta or Final. Or we can just push the candidate to internal artifactory but not upload it to maven until its stable enough for general use.

willsalz:
Merging to unblock.

PR #78, opened 2015-09-23
hyperbahn client first pass

### LINKED ISSUE #77: InitRequestInitiator should send current host:port if server is listening
So, this bit us because Hyperbahn uses the host:port value set in the original InitRequest message instead a value passed in as part of a Advertise request to determine where to reach the service. Needless to say I'll be filing a bug on that, but, as that issue is unlikely to get any traction, we'll need to patch our implementation ( and, in doing so, get up to spec with the protocol ).

The implications here are that ChannelInitializers for outbounds e.g. client requests will need to modify their ChannelPipeline to pass in the listening port of their owning TChannel instance. This is tricker than it sound due to the order of operations of how Pipelines are created ( typically pipelines are created prior to  listening but we'll have to modify client pipelines after a server pipeline is created )

willsalz:
What
====
Really basic Hyperbahn Client. Not yet 100% functioning due to TChannel bug #77 

Why
====
because hyperbahn. 

Misc
====
- Actually set the TTL on an outbound, oops.
- return an empty header map if Gson returns null on parse ( that would happen if there were no headers )
- Add a basic hyperbahn example
- fix some aggressive renaming of example request handlers
- implement basic hyperbahn client

Todo
====
- Fix #77 
- Perform a heartbeat with the Hyperbahn routers so we don't get disconnected
- implement the exponential backoff strategy for connecting

cc @truncs @brandoniles @junchaowu

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
@@ -29,6 +29,7 @@ import io.netty.buffer.Unpooled;
 import io.netty.util.CharsetUtil;
 
 import java.lang.reflect.Type;
+import java.util.HashMap;
 import java.util.Map;
 
 public final class JSONSerializer implements Serializer.SerializerInterface {
@@ -47,7 +48,8 @@ public final class JSONSerializer implements Serializer.SerializerInterface {
     public Map<String, String> decodeHeaders(ByteBuf arg2) {
         String headerJSON = arg2.toString(CharsetUtil.UTF_8);
         arg2.release();
-        return new Gson().fromJson(headerJSON, HEADER_TYPE);
+        Map<String, String> headers = new Gson().fromJson(headerJSON, HEADER_TYPE);
+        return (headers == null) ? new HashMap<String, String>() : headers;
     }
 
     @Override
--- a/tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
+++ b/tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
@@ -27,17 +27,12 @@ import com.uber.tchannel.api.TChannel;
 import com.uber.tchannel.hyperbahn.api.HyperbahnClient;
 import com.uber.tchannel.hyperbahn.messages.AdvertiseResponse;
 
-import java.net.UnknownHostException;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeoutException;
-
 public class HyperbahnExample {
-    public static void main(String[] args) throws UnknownHostException, InterruptedException, ExecutionException,
-            TimeoutException {
+    public static void main(String[] args) throws Exception {
         TChannel tchannel = new TChannel.Builder("hyperbahn-example").build();
         HyperbahnClient hyperbahn = new HyperbahnClient(tchannel);
         Response<AdvertiseResponse> response = hyperbahn.advertise(tchannel.getServiceName(), 0);
         System.err.println(response);
-
+        hyperbahn.shutdown();
     }
 }
--- a/tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
+++ b/tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
@@ -26,18 +26,26 @@ import com.uber.tchannel.api.Response;
 import com.uber.tchannel.api.TChannel;
 import com.uber.tchannel.hyperbahn.api.HyperbahnClient;
 import com.uber.tchannel.hyperbahn.messages.AdvertiseResponse;
-import io.netty.channel.ChannelFuture;
+import com.uber.tchannel.ping.PingRequestHandler;
 
 import java.util.concurrent.TimeUnit;
 
 public class HyperbahnExample {
     public static void main(String[] args) throws Exception {
-        TChannel tchannel = new TChannel.Builder("hyperbahn-example").build();
-        ChannelFuture f = tchannel.listen();
-        final HyperbahnClient hyperbahn = new HyperbahnClient(tchannel);
+        TChannel tchannel = new TChannel.Builder("hyperbahn-example")
+                .register("ping", new PingRequestHandler())
+                .build();
+
+        tchannel.listen();
+
+        HyperbahnClient hyperbahn = new HyperbahnClient(tchannel);
+
         Response<AdvertiseResponse> response = hyperbahn.advertise(tchannel.getServiceName(), 0);
+
         System.err.println(response);
+
         Thread.sleep(TimeUnit.MILLISECONDS.convert(1, TimeUnit.MINUTES));
+
         hyperbahn.shutdown();
     }
 }
--- a/tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
+++ b/tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
@@ -26,13 +26,18 @@ import com.uber.tchannel.api.Response;
 import com.uber.tchannel.api.TChannel;
 import com.uber.tchannel.hyperbahn.api.HyperbahnClient;
 import com.uber.tchannel.hyperbahn.messages.AdvertiseResponse;
+import io.netty.channel.ChannelFuture;
+
+import java.util.concurrent.TimeUnit;
 
 public class HyperbahnExample {
     public static void main(String[] args) throws Exception {
         TChannel tchannel = new TChannel.Builder("hyperbahn-example").build();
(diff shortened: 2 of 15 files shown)


tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
```diff
@@ -47,7 +48,8 @@ public String decodeEndpoint(ByteBuf arg1) {
     public Map<String, String> decodeHeaders(ByteBuf arg2) {
         String headerJSON = arg2.toString(CharsetUtil.UTF_8);
         arg2.release();
-        return new Gson().fromJson(headerJSON, HEADER_TYPE);
+        Map<String, String> headers = new Gson().fromJson(headerJSON, HEADER_TYPE);
```
truncs: this looks a bit weird, won't the headerJSON be null ? or its just {} ?

tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
```diff
@@ -0,0 +1,51 @@
...
+    public static void main(String[] args) throws Exception {
+        TChannel tchannel = new TChannel.Builder("hyperbahn-example")
+                .register("ping", new PingRequestHandler())
+                .build();
+
+        tchannel.listen();
+
+        HyperbahnClient hyperbahn = new HyperbahnClient(tchannel);
+
+        Response<AdvertiseResponse> response = hyperbahn.advertise(tchannel.getServiceName(), 0);
```
truncs: Ideally the end developer should really care about the response coming back from the advertise message ?

tchannel-example/src/main/java/com/uber/tchannel/json/JsonRequestHandler.java
```diff
@@ -27,7 +27,7 @@
 import com.uber.tchannel.api.Response;
 import com.uber.tchannel.api.ResponseCode;
 
-public class JsonDefaultRequestHandler extends DefaultRequestHandler<RequestPojo, ResponsePojo> {
+public class JsonRequestHandler extends DefaultRequestHandler<RequestPojo, ResponsePojo> {
```
truncs: you are going to break my merge

tchannel-example/src/main/java/com/uber/tchannel/thrift/GetValueRequestHandler.java
```diff
@@ -31,11 +31,11 @@
 
 import java.util.Map;
 
-public class GetValueHandlerDefault extends DefaultRequestHandler<KeyValue.getValue_args, KeyValue.getValue_result> {
+public class GetValueRequestHandler extends DefaultRequestHandler<KeyValue.getValue_args, KeyValue.getValue_result> {
```
truncs: ditto, about breaking the merge, we can move the renaming changes in a different PR ?

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.regex.Pattern;
+
+public class HyperbahnClient {
+
+    private static final String HOSTS_FILE_PATH = "/etc/uber/hyperbahn/hosts.json";
```
truncs: this makes little sense for OSS, internal developers should pass the file path manually even if it means some boilerplate code.

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
```diff
@@ -0,0 +1,123 @@
...
+    private final TChannel tchannel;
+    private final Logger logger = LoggerFactory.getLogger(HyperbahnClient.class);
+    private List<InetSocketAddress> routers;
+
+    public HyperbahnClient(TChannel tchannel) throws IOException {
+        this.tchannel = tchannel;
+        this.routers = HyperbahnClient.loadRouters();
+    }
+
+    private static List<InetSocketAddress> loadRouters() throws IOException {
```
truncs: this function is weird, its a static function with sideeffects to routers.

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
```diff
@@ -0,0 +1,123 @@
...
+            int port = Integer.parseInt(hostPortPair[1]);
+            InetSocketAddress router = new InetSocketAddress(host, port);
+            routers.add(router);
+        }
+
+        return routers;
+
+    }
+
+    public Response<AdvertiseResponse> advertise(
```
truncs: Advertise needs to be a continuous thing, I guess you are planning to add that later ?

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
```diff
@@ -0,0 +1,123 @@
...
+        final InetSocketAddress router = this.routers.get(new Random().nextInt(this.routers.size()));
+
+        Future<Response<AdvertiseResponse>> responseFuture = this.tchannel.callJSON(
+                router.getAddress(),
+                router.getPort(),
+                request,
+                AdvertiseResponse.class
+        );
+
+        return responseFuture.get(2, TimeUnit.SECONDS);
```
truncs: this makes the advertise message blocking.

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
```diff
@@ -0,0 +1,123 @@
...
+
+    }
+
+    public void stopAdvertising() {
+
+    }
+
+    public void shutdown() throws InterruptedException {
+        this.logger.info("Shutting down HyperbahnClient and TChannel.");
```
truncs: nitpick: just logger.info ?

tchannel-hyperbahn/src/test/java/com/uber/tchannel/hyperbahn/api/HyperbahnClientTest.java
```diff
@@ -0,0 +1,36 @@
...
+ * THE SOFTWARE.
+ */
+
+package com.uber.tchannel.hyperbahn.api;
+
+import com.uber.tchannel.api.TChannel;
+
+public class HyperbahnClientTest {
+
+    @org.junit.Test
```
truncs: why not import org.junit.Test
and then just use 
@Test

tchannel-hyperbahn/src/test/java/com/uber/tchannel/hyperbahn/api/HyperbahnClientTest.java
```diff
@@ -0,0 +1,36 @@
...
+    public void testAdvertise() throws Exception {
+        TChannel tchannel = new TChannel.Builder("hyperbahn-service").build();
+
+        HyperbahnClient hyperbahnClient = new HyperbahnClient(tchannel);
```
truncs: WIP ?

tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
willsalz: It can be a variety of things, and it depends on the caller. This prevents a NPE when Gson can't decode invalid json e.g. empty strings or nulls.

tchannel-example/src/main/java/com/uber/tchannel/json/JsonRequestHandler.java
willsalz: shouldn't have renamed all the example handlers then :)

willsalz: but yeah can rename in a different PR.

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
willsalz: yup makes sense.

tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java
willsalz: yup, this is just a WIP

willsalz:
Going to break this in to a few diffs, one for bug fixes, one for renaming, one for the actual hyperbahn client.

PR #80, opened 2015-10-02
Prevent NPE when clients send null JSON headers

willsalz:
What
-------
Two things:
- return an empty HashMap if the TChannel-as-JSON request comes in with null headers ( as Gson will just return `null` ).
- `Gson` objects are thread safe ( not that we're using them across threads anyway ) so move the codec to a static member and just use that for 'performance'.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
@@ -29,12 +29,15 @@ import io.netty.buffer.Unpooled;
 import io.netty.util.CharsetUtil;
 
 import java.lang.reflect.Type;
+import java.util.HashMap;
 import java.util.Map;
 
 public final class JSONSerializer implements Serializer.SerializerInterface {
     private static final Type HEADER_TYPE = (new TypeToken<Map<String, String>>() {
 
     }).getType();
+
+    private static final Gson GSON = new Gson();
 
     @Override
     public String decodeEndpoint(ByteBuf arg1) {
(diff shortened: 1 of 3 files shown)


tchannel-core/src/main/java/com/uber/tchannel/schemes/JSONSerializer.java
```diff
@@ -47,14 +50,15 @@ public String decodeEndpoint(ByteBuf arg1) {
     public Map<String, String> decodeHeaders(ByteBuf arg2) {
         String headerJSON = arg2.toString(CharsetUtil.UTF_8);
         arg2.release();
-        return new Gson().fromJson(headerJSON, HEADER_TYPE);
+        Map<String, String> headers = GSON.fromJson(headerJSON, HEADER_TYPE);
+        return (headers != null) ? headers : new HashMap<String, String>();
```
truncs: can the same thing happen with Thrift as well ?

willsalz: yes probably. I'll go over the serializers and make sure that they are sane.

truncs: Also what happens on the encoding side during null, should we use EmptyByteBuf instead of nulls since the underlying network stack assumes that is always going to be a bytebuf ?

willsalz:
@truncs mind if I land? There will be an upcoming diff that I don't want to land yet to handle ( undocumented 😑 ) changes to the JSON headers.

truncs:
@willsalz squash the commits and then land ?

PR #83, opened 2015-10-14
support as-raw requests

ShanniLi:
r @willsalz @truncs

a simple start. :)

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/schemes/RawSerializer.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/schemes/RawSerializer.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2015 Uber Technologies, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


truncs:
@ShanniLi Raw requests are already supported, the only thing that is required is use the call method with a RawRequest and it would return a RawResponse. Infact I am benchmarking right now with a RawRequest and RawResponses.

tchannel-core/src/main/java/com/uber/tchannel/schemes/RawSerializer.java
```diff
@@ -0,0 +1,90 @@
...
+    private static final Gson GSON = new Gson();
+
+    @Override
+    public String decodeEndpoint(ByteBuf arg1) {
+        String endpoint = arg1.toString(CharsetUtil.UTF_8);
+        arg1.release();
+        return endpoint;
+    }
+
+//    @Override
```
willsalz: nit: don't commit commented code

tchannel-core/src/main/java/com/uber/tchannel/schemes/RawSerializer.java
```diff
@@ -0,0 +1,90 @@
...
+        arg3.release();
+        return body;
+    }
+
+    @Override
+    public ByteBuf encodeEndpoint(String method) {
+        return Unpooled.wrappedBuffer(method.getBytes());
+    }
+
+//    @Override
```
willsalz: same here.

tchannel-core/src/main/java/com/uber/tchannel/schemes/RawSerializer.java
```diff
@@ -0,0 +1,90 @@
...
+//    public <T> T decodeHeaders(ByteBuf arg2) {
+//        String header = arg2.toString(CharsetUtil.UTF_8);
+//        arg2.release();
+//        return (T)header;
+//    }
+
+    @Override
+    public Map<String, String> decodeHeaders(ByteBuf arg2) {
```
willsalz: This seems incorrect; why are you decoding raw args at all?

tchannel-core/src/main/java/com/uber/tchannel/schemes/RawSerializer.java
```diff
@@ -0,0 +1,90 @@
...
+        String headerJSON = arg2.toString(CharsetUtil.UTF_8);
+        arg2.release();
+        Map<String, String> headers = GSON.fromJson(headerJSON, HEADER_TYPE);
+        return (headers != null) ? headers : new HashMap<String, String>();
+    }
+
+    @Override
+    public <T> T decodeBody(ByteBuf arg3, Class<T> bodyType) {
```
willsalz: Again, the Raw Arg Scheme makes no assumptions about the body; so why is this being decoded as a string? And subsequently cast to a `T`?

To make the downstream consumer's life easy, we should be returning `byte[]`, then `releasing()` the reference counted buffer to make sure that the client isn't responsible for that.

ShanniLi:
@truncs, the reason for this RP is to provide a consistent interface for as-raw like others. as-raw is actually heavily used ...
@willsalz very good points! I will put this PR to shelf first. I will need to understand more about as-thrift to make a better version of this PR.

ref: #84

PR #85, opened 2015-10-16
Protocol v2, init handshake, init headers, etc.

ShanniLi:
Efforts to make tchannel-java talking to other languages & Hyperbahn:
1. Init req/res handshake should finish before other req/res can start: tchannel v2 protocol
2. Init req/res should have the host_port set when listen. It should use "0.0.0.0:0" otherwise for ephemeral ports.
3. Connection pool and peer selection support, e.g., should be able to prefer Hyperbahn affinity connections. 

r: @willsalz @truncs @anson627 

I feel this is enough for the first step to unblock Hyperbahn client & talking to other languages. The are still many places to improve. E.g., the connection pool can be a bottleneck for multi-threading. It can be relieved by read-write locks. But all for next steps.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -166,7 +166,7 @@ public final class TChannel {
         InetSocketAddress localAddress = (InetSocketAddress) f.channel().localAddress();
         this.listeningPort = localAddress.getPort();
         this.listeningHost = localAddress.getHostName();
-        this.peerManager.setHostPort(this.listeningHost, this.listeningPort);
+        this.peerManager.setHostPort(String.format("%s:%d", this.listeningHost, this.listeningPort));
         return f;
     }
 
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -161,6 +161,10 @@ public final class TChannel {
         return this.service;
     }
 
+    public PeerManager getPeerManager() {
+        return this.peerManager;
+    }
+
     public ChannelFuture listen() throws InterruptedException {
         ChannelFuture f = this.serverBootstrap.bind(this.host, this.port).sync();
         InetSocketAddress localAddress = (InetSocketAddress) f.channel().localAddress();
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -24,6 +24,8 @@ package com.uber.tchannel.api;
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.SettableFuture;
+import com.uber.tchannel.api.errors.TChannelError;
+import com.uber.tchannel.api.errors.TChannelConnectionTimeout;
 import com.uber.tchannel.api.handlers.RequestHandler;
 import com.uber.tchannel.channels.ChannelManager;
 import com.uber.tchannel.channels.ChannelRegistrar;
@@ -78,9 +80,11 @@ public final class TChannel {
     private final EventLoopGroup childGroup;
     private final InetAddress host;
     private final int port;
+    private String listeningHost = "0.0.0.0";
     private int listeningPort;
     private ExecutorService exectorService;
     private final int maxQueuedRequests;
+    private final int initTimeout;
 
     private final Serializer serializer = new Serializer(new HashMap<ArgScheme, Serializer.SerializerInterface>() {
         {
@@ -100,6 +104,7 @@ public final class TChannel {
         this.host = builder.host;
         this.port = builder.port;
         this.maxQueuedRequests = builder.maxQueuedRequests;
+        this.initTimeout = builder.initTimeout;
     }
 
     private <T, U> ListenableFuture<Response<T>> callWithEncoding(
@@ -108,7 +113,7 @@ public final class TChannel {
             Request<U> request,
             final Class<T> responseType,
             ArgScheme scheme
-    ) throws InterruptedException {
+    ) throws InterruptedException, TChannelError {
 
         RawRequest rawRequest = new RawRequest(
                 request.getTTL(),
@@ -160,6 +165,8 @@ public final class TChannel {
         ChannelFuture f = this.serverBootstrap.bind(this.host, this.port).sync();
         InetSocketAddress localAddress = (InetSocketAddress) f.channel().localAddress();
         this.listeningPort = localAddress.getPort();
+        this.listeningHost = localAddress.getHostName();
+        this.channelManager.setHostPort(this.listeningHost, this.listeningPort);
         return f;
     }
 
@@ -174,7 +181,7 @@ public final class TChannel {
             int port,
             Request<U> request,
             final Class<T> responseType
-    ) throws InterruptedException {
+    ) throws InterruptedException, TChannelError {
         return callWithEncoding(host, port, request, responseType, ArgScheme.THRIFT);
     }
 
@@ -183,7 +190,7 @@ public final class TChannel {
             int port,
             Request<U> request,
             final Class<T> responseType
-    ) throws InterruptedException {
+    ) throws InterruptedException, TChannelError {
         return callWithEncoding(host, port, request, responseType, ArgScheme.JSON);
     }
 
@@ -191,7 +198,7 @@ public final class TChannel {
             InetAddress host,
             int port,
             RawRequest request
-    ) throws InterruptedException {
+    ) throws InterruptedException, TChannelError {
 
         // Set the ArgScheme as RAW if its not set
         Map<String, String> transportHeaders = request.getTransportHeaders();
@@ -200,7 +207,11 @@ public final class TChannel {
         }
 
         // Get an outbound channel
-        Channel ch = this.channelManager.findOrNew(new InetSocketAddress(host, port), this.clientBootstrap);
+        Channel ch = this.channelManager.findOrNew(new InetSocketAddress(host, port), this.clientBootstrap).channel;
+
+        if (!this.channelManager.waitForIdentified(ch, this.initTimeout)) {
+            throw new TChannelConnectionTimeout();
+        }
 
         // Get a response router for our outbound channel
         ResponseRouter responseRouter = ch.pipeline().get(ResponseRouter.class);
@@ -221,6 +232,7 @@ public final class TChannel {
         private EventLoopGroup bossGroup = new NioEventLoopGroup(1);
         private EventLoopGroup childGroup = new NioEventLoopGroup();
         private LogLevel logLevel = LogLevel.INFO;
+        private int initTimeout = 2000;
 
         public Builder(String service) throws UnknownHostException {
             if (service == null) {
@@ -270,6 +282,11 @@ public final class TChannel {
             return this;
         }
 
+        public Builder setInitTimeout(int initTimeout) {
+            this.initTimeout = initTimeout;
+            return this;
+        }
+
         public TChannel build() {
             return new TChannel(this);
         }
@@ -307,9 +324,9 @@ public final class TChannel {
                     ch.pipeline().addLast("MessageCodec", new MessageCodec());
 
                     if (isServer) {
-                        ch.pipeline().addLast("InitRequestHandler", new InitRequestHandler());
+                        ch.pipeline().addLast("InitRequestHandler", new InitRequestHandler(channelManager));
                     } else {
-                        ch.pipeline().addLast("InitRequestInitiator", new InitRequestInitiator());
+                        ch.pipeline().addLast("InitRequestInitiator", new InitRequestInitiator(channelManager));
                     }
 
                     // Handle PingRequest
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -27,7 +27,7 @@ import com.google.common.util.concurrent.SettableFuture;
 import com.uber.tchannel.api.errors.TChannelError;
 import com.uber.tchannel.api.errors.TChannelConnectionTimeout;
 import com.uber.tchannel.api.handlers.RequestHandler;
-import com.uber.tchannel.channels.ChannelManager;
+import com.uber.tchannel.channels.PeerManager;
 import com.uber.tchannel.channels.ChannelRegistrar;
 import com.uber.tchannel.codecs.MessageCodec;
 import com.uber.tchannel.codecs.TChannelLengthFieldBasedFrameDecoder;
@@ -75,7 +75,7 @@ public final class TChannel {
     private final String service;
     private final ServerBootstrap serverBootstrap;
     private final Bootstrap clientBootstrap;
-    private final ChannelManager channelManager;
+    private final PeerManager peerManager;
     private final EventLoopGroup bossGroup;
     private final EventLoopGroup childGroup;
     private final InetAddress host;
@@ -98,7 +98,7 @@ public final class TChannel {
         this.exectorService = builder.executorService;
         this.serverBootstrap = builder.serverBootstrap();
         this.clientBootstrap = builder.bootstrap();
-        this.channelManager = builder.channelManager;
+        this.peerManager = builder.peerManager;
         this.bossGroup = builder.bossGroup;
         this.childGroup = builder.childGroup;
         this.host = builder.host;
@@ -166,12 +166,12 @@ public final class TChannel {
         InetSocketAddress localAddress = (InetSocketAddress) f.channel().localAddress();
         this.listeningPort = localAddress.getPort();
         this.listeningHost = localAddress.getHostName();
-        this.channelManager.setHostPort(this.listeningHost, this.listeningPort);
+        this.peerManager.setHostPort(this.listeningHost, this.listeningPort);
         return f;
     }
 
     public void shutdown() throws InterruptedException {
-        this.channelManager.close();
+        this.peerManager.close();
         this.bossGroup.shutdownGracefully();
         this.childGroup.shutdownGracefully();
     }
@@ -207,9 +207,9 @@ public final class TChannel {
         }
 
         // Get an outbound channel
-        Channel ch = this.channelManager.findOrNew(new InetSocketAddress(host, port), this.clientBootstrap).channel;
+        Channel ch = this.peerManager.findOrNew(new InetSocketAddress(host, port), this.clientBootstrap).channel();
 
-        if (!this.channelManager.waitForIdentified(ch, this.initTimeout)) {
+        if (!this.peerManager.waitForIdentified(ch, this.initTimeout)) {
             throw new TChannelConnectionTimeout();
         }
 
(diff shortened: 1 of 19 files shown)


tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
```diff
@@ -41,47 +39,102 @@
...
+    private final Map<SocketAddress, Peer> peers = new Hashtable<>();
+    private String host = "0.0.0.0";
+    private int port = 0;
 
-    private final Map<InetSocketAddress, ChannelId> channelMap = new HashMap<>();
-    private final ChannelGroup channels = new DefaultChannelGroup(GlobalEventExecutor.INSTANCE);
+    public Connection findOrNew(SocketAddress address, Bootstrap bootstrap) throws InterruptedException {
+        Peer peer = peers.get(address);
+        if (peer == null) {
+            synchronized (peers) {
```
willsalz: can you do this without locking?

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
```diff
@@ -41,47 +39,102 @@
  * entry point for shutting down all active Channels.
  */
 public class ChannelManager {
```
willsalz: If all of these methods are returning Connections, how about renaming this ConnectionManager instead? Seems misleading otherwise.

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.util.Map;
+
+/**
+ * Connection represents a connection to a remote address
+ */
+public class Connection {
+    public Channel channel;
+    public String remoteAddress = null;
+    public Direction direction = Direction.NONE;
```
willsalz: Why does direction matter?

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
```diff
@@ -41,47 +39,102 @@
...
-        return channel;
+        return peer.handleActiveConnection(ctx, getDirection(ctx)).channel;
     }
 
-    public void close() throws InterruptedException {
-        this.channels.close().sync();
-        this.channels.clear();
-        this.channelMap.clear();
+    public static Connection.Direction getDirection(ChannelHandlerContext ctx) {
+        if (ctx.pipeline().names().contains("InitRequestHandler")) {
```
willsalz: The `InitRequestHandler` removes itself from the pipeline after the handshake so this logic may be incorrect.

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
```diff
@@ -41,47 +39,102 @@
...
+        if (peer != null) {
+            return peer.getConnection(channel.id());
+        }
+
+        return null;
     }
 
-    public Channel findOrNew(InetSocketAddress address, Bootstrap bootstrap) throws InterruptedException {
-        Channel channel = this.find(address);
+    public Channel add(ChannelHandlerContext ctx) {
```
willsalz: why do other methods take `Channel` but this take a ctx?

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
```diff
@@ -0,0 +1,116 @@
...
+import com.uber.tchannel.messages.InitMessage;
+import io.netty.channel.Channel;
+
+import java.util.Map;
+
+/**
+ * Connection represents a connection to a remote address
+ */
+public class Connection {
+    public Channel channel;
```
willsalz: nit: make private and immutable.

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
```diff
@@ -0,0 +1,116 @@
...
+    public String remoteAddress = null;
```
willsalz: nit: private and immutable

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
willsalz: nit: private and immutable.

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
```diff
@@ -0,0 +1,116 @@
...
+
+        this.remoteAddress = hostPort.trim();
+        this.setState(ConnectionState.IDENTIFIED);
+    }
+
+    public synchronized boolean isEphemeral() {
+        return this.remoteAddress.equals("0.0.0.0:0");
+    }
+
+    public static String[] sliptHostPort(String hostPort) {
```
willsalz: typo

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -0,0 +1,145 @@
...
+import com.uber.tchannel.messages.InitRequest;
+import io.netty.bootstrap.Bootstrap;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelId;
+import io.netty.channel.ChannelFuture;
+import io.netty.channel.ChannelFutureListener;
+import io.netty.channel.ChannelHandlerContext;
+
+/**
+ * Peer keeps track of open channels, and provides a way to lookup Channels by their Remote Address.
```
willsalz: This is what the ChannelManager already did…

tchannel-core/src/main/java/com/uber/tchannel/handlers/InitRequestHandler.java
```diff
@@ -64,6 +76,7 @@ public void messageReceived(ChannelHandlerContext ctx, Message message) throws E
 
             default:
 
+                // TODO: should send back BadRequest
```
willsalz: I'm not so sure. Bad Request is defined as `Request args do not match expectations, request will never be satisfied, do not retry.`  InitRequest doesn't even have `args`.

willsalz:
Also your builds are failing.

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
ShanniLi: No, netty has a thread pool for async callbacks. There will be race conditions without lock.

anson627: you don't have to lock the new and connect part. Instead of put, maybe change Hashtable to concurrenthashmap, and use putIfAbsent?

junchaowu: +1

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -270,6 +282,11 @@ public Builder setLogLevel(LogLevel logLevel) {
             return this;
         }
 
+        public Builder setInitTimeout(int initTimeout) {
+            this.initTimeout = initTimeout;
+            return this;
```
junchaowu: what's the purpose return itself.

willsalz: this enables a 'fluent interface' e.g. builder chaining.

tchannel-core/src/main/java/com/uber/tchannel/api/errors/TChannelError.java
```diff
@@ -0,0 +1,42 @@
...
+public class TChannelError extends Exception {
+    public static final String ERROR_INIT_TIMEOUT = "tchannel.connection.timeout";
+
+    public final String type;
+    public final Exception subError;
+
+    public TChannelError(String message, String type, Exception subError) {
+        super(message);
+        this.type = type;
+        this.subError = subError;
```
junchaowu: what does subError mean

tchannel-core/src/main/java/com/uber/tchannel/api/errors/TChannelError.java
```diff
@@ -0,0 +1,42 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.tchannel.api.errors;
+
+public class TChannelError extends Exception {
```
willsalz: Is this representing an actual error as defined in the Protocol? If so there is already an Class structure that you should interred from, namely the `ProtocolError` class

tchannel-core/src/main/java/com/uber/tchannel/handlers/InitRequestHandler.java
```diff
@@ -46,14 +53,19 @@ public void messageReceived(ChannelHandlerContext ctx, Message message) throws E
...
                 if (initRequestMessage.getVersion() == InitMessage.DEFAULT_VERSION) {
-                    ChannelFuture f = ctx.writeAndFlush(new InitResponse(
+                    InitResponse initResponse = new InitResponse(
                             initRequestMessage.getId(),
-                            InitMessage.DEFAULT_VERSION,
-                            initRequestMessage.getHeaders()
-                    ));
+                            InitMessage.DEFAULT_VERSION
+                    );
+                    initResponse.setHostPort(this.channelManager.getHostPort());
```
willsalz: Why add mutating methods? Just put this in the constructor.

tchannel-core/src/main/java/com/uber/tchannel/handlers/InitRequestHandler.java
```diff
@@ -46,14 +53,19 @@ public void messageReceived(ChannelHandlerContext ctx, Message message) throws E
...
+                    // TODO: figure out what to put here
+                    initResponse.setProcessName("java-process");
```
willsalz: should be the 'service name' set in the `TChannel` top-level object, no?

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
ShanniLi: Good suggestions! I will create a task for this optimization. The goal of this PR and the next several  PRs is to bump up tchannel-java to be able to talk to other languages & Hyperbahn. Let's get it working before optimizing the perf. :)

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
ShanniLi: Direction matters when talking to Hyperbahn, since affinity connections are always IN. We want all clients to use affinity connections to talk to Hyperbahn.

The current code also has a bug that leaks connections when it has an OUT connection to the host and later has an IN connection back from the same host.

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
ShanniLi: Good point! I will inline the code to the caller which will only be executed for the first time.

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
ShanniLi: If it is an out connection, it needs to send out the init request.

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
ShanniLi: yea, that is because I copied the comment from ChannelManager ... I will update the it.

tchannel-core/src/main/java/com/uber/tchannel/handlers/InitRequestHandler.java
ShanniLi: yes, should be ProtocolError. Exception should not be thrown here. TChannel should just send back protocol error and reset the connection. All for future PRs ...

tchannel-core/src/main/java/com/uber/tchannel/api/errors/TChannelError.java
ShanniLi: TChannelError can be an error wrapper for other issues. This is for that.

tchannel-core/src/main/java/com/uber/tchannel/api/errors/TChannelError.java
ShanniLi: No, the ProtocolError is for TChannel error frames. This is for all TChannel errors, e.g., connection timeout, connection reset, response already done, etc.

Another point, ProtocolError is one type of TChannel error frame, so it may not be a good idea to make it a base class. But this would be for a different adventure to refactor it.

tchannel-core/src/main/java/com/uber/tchannel/handlers/InitRequestHandler.java
ShanniLi: I want to be able to change it in test.

tchannel-core/src/main/java/com/uber/tchannel/handlers/InitRequestHandler.java
ShanniLi: I think this should be the name of the process that sends the request. E.g., JsonClient ...

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
ShanniLi: issue created to track it: #86

tchannel-core/src/main/java/com/uber/tchannel/channels/ChannelManager.java
ShanniLi: It is really PeerManager. I will rename it to that.

tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
ShanniLi: Cannot, remoteAddress is only initialized when the connection gets identified.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -108,7 +113,7 @@ private TChannel(Builder builder) {
             Request<U> request,
             final Class<T> responseType,
             ArgScheme scheme
-    ) throws InterruptedException {
+    ) throws InterruptedException, TChannelError {
```
truncs: what does a TChannelError suppose to represent over here ?

ShanniLi: https://github.com/uber/tchannel-node/blob/master/errors.js

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -0,0 +1,146 @@
...
+    public SocketAddress remoteAddress = null;
+
+    private PeerManager manager;
+
+    public Peer(PeerManager manager, SocketAddress remoteAddress) {
+        this.manager = manager;
+        this.remoteAddress = remoteAddress;
+    }
+
+    public synchronized Connection add(Connection connection) {
```
truncs: why not use a ConcurrentHashMap instead of making all these functions synchronized ?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
truncs: so its a SuperClass for all TChannel Errors ?

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -0,0 +1,146 @@
...
+        Connection conn = maps.get(channel.id());
+        if (conn == null) {
+            conn = new Connection(channel, direction);
+            add(conn);
+        }
+
+        return conn;
+    }
+
+    public synchronized Connection handleActiveConnection(ChannelHandlerContext ctx, Connection.Direction direction) {
```
truncs: why is this class not inheriting from ChannelHandler ? I don't think we should be passing the ChannelHandlerContext outside Netty

tchannel-core/src/main/java/com/uber/tchannel/channels/PeerManager.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+public class PeerManager {
+    private final Map<SocketAddress, Peer> peers = new Hashtable<>();
+    private String host = "0.0.0.0";
+    private int port = 0;
+
+    public Connection findOrNew(SocketAddress address, Bootstrap bootstrap) throws InterruptedException {
+        Peer peer = peers.get(address);
+        if (peer == null) {
+            synchronized (peers) {
```
truncs: just use ConcurentHashMap ?

tchannel-core/src/test/java/com/uber/tchannel/handlers/InitDefaultRequestHandlerTest.java
```diff
@@ -51,7 +52,7 @@ public void testInitHandlerRemovesItself() throws Exception {
 
         // Given
         EmbeddedChannel channel = new EmbeddedChannel(
```
truncs: I don't see much test cases around the new classes ?

truncs:
@ShanniLi Few questions

1) Why can't the peer affinity logic live in the the hyperbahn module ?
2) Have you tested this with a Hyperbahn cluster ?
3) I am doing some benchmarking without this diff and its going to affect my work, how can we avoid it by making minimal changes and also to unblock hyperbahn ?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: yes

tchannel-core/src/test/java/com/uber/tchannel/handlers/InitDefaultRequestHandlerTest.java
ShanniLi: The new classes are mostly changes of the backend logic. Most of the current test should cover the correct behaviors. I have also updated the init test to validate the headers.

I will add some validations in the code logic to make sure a req can never be sent without a identification; however it will not be in this PR.

I will also add tests for peer/connection maintenance and clean up, but that wouldn't be necessary until the peer/connection reaping logic is in.

ShanniLi:
I have update the code to use ConcurrentHashMap per suggestions from @anson627 and @truncs

ShanniLi:
@truncs 
1. Peer affinity logic lives in Hyperbahn. However, the client also needs to honor that.
2. This is just the very first step to get tchannel-java to talk in tchannel-v2. Talking to Hyperbahn will be the next step (may be several steps later depends on what's the next problem I find).
3. I think it would be better to do benchmark based on the latest tchannel protocol. What tchannel-java talks right now isn't really a tchannel protocol other languages (node, python, and go) respect. Let's makes sure it works at the protocol level before diving in to the perf work. I can help rebase your branch after the critical features/fixes are in.

anson627:
lgtm

PR #89, opened 2015-10-22
Sub-channel, peer selection, error frame handling, etc.

ShanniLi:
r @anson627 @willsalz @truncs

Another major push to tchannel-v2:
1. subChannel: allow singleton & connection pooling
2. Peer selection: necessary for retry & traffic balancing
3. Error frame handling: required by tchannel-v1


tchannel-example/src/main/java/com/uber/tchannel/thrift/KeyValueClient.java
```diff
@@ -82,12 +84,14 @@ public static String getValue(
 
         KeyValue.getValue_args getValue = new KeyValue.getValue_args(key);
 
-        ListenableFuture<Response<KeyValue.getValue_result>> future = tchannel.callThrift(
+        ListenableFuture<Response<KeyValue.getValue_result>> future = tchannel
+            .makeSubChannel("keyvalue-service")
+            .callThrift(
                 InetAddress.getLocalHost(),
```
anson627: set host port should belong to subChannel as well?

anson627:
channel should worry about hostname and port, the callJson/callThrift should not care.

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
```diff
@@ -0,0 +1,261 @@
...
+        return requestHandlers.get(endpoint);
+    }
+
+    public Connection.Direction getPreferredDirection() {
+        return preferredDirection;
+    }
+
+    public SubChannel setPeers(List<InetSocketAddress> peers) {
+        for (InetSocketAddress peer : peers) {
+            this.peers.add(new SubPeer(peer, this));
```
willsalz: circular reference, need a weak reference?

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
```diff
@@ -0,0 +1,261 @@
...
+    public <T, U> ListenableFuture<Response<T>> callJSON(
+            InetAddress host,
+            int port,
+            Request<U> request,
+            final Class<T> responseType
+    ) throws InterruptedException, TChannelError {
+        return callWithEncoding(host, port, request, responseType, ArgScheme.JSON);
+    }
+
+    public <T, U> ListenableFuture<Response<T>> callJSON(
```
willsalz: what is this? calling a null host?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -174,69 +129,38 @@ public ChannelFuture listen() throws InterruptedException {
...
-
-        // Set the ArgScheme as RAW if its not set
-        Map<String, String> transportHeaders = request.getTransportHeaders();
-        if (!transportHeaders.containsKey(TransportHeaders.ARG_SCHEME_KEY)) {
-            request.setTransportHeader(TransportHeaders.ARG_SCHEME_KEY, ArgScheme.RAW.getScheme());
+    // TODO: make sure makeSubChannel is called before listening ...
+    public SubChannel makeSubChannel(String service, Connection.Direction preferredDirection) {
+        SubChannel subChannel = getSubChannel(service);
+        if (subChannel == null) {
+            subChannel = new SubChannel(service, this, preferredDirection);
```
willsalz: more circular references.

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -29,8 +29,8 @@
 
 import com.uber.tchannel.api.errors.TChannelConnectionFailure;
 import com.uber.tchannel.api.errors.TChannelError;
-import com.uber.tchannel.messages.InitMessage;
-import com.uber.tchannel.messages.InitRequest;
+import com.uber.tchannel.frames.InitFrame;
```
willsalz: that's not what that message is called…

tchannel-core/src/main/java/com/uber/tchannel/checksum/Checksums.java
```diff
@@ -21,25 +21,25 @@
  */
 package com.uber.tchannel.checksum;
 
-import com.uber.tchannel.messages.CallMessage;
+import com.uber.tchannel.frames.CallFrame;
```
willsalz: why are you renaming all of the messages? If you want to do that put it in its own diff.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
anson627: java GC is smart, so circular is fine.

tchannel-core/src/main/java/com/uber/tchannel/frames/ErrorFrame.java
```diff
@@ -76,4 +80,19 @@ public String toString() {
         );
     }
 
+    public static void sendError(ErrorType type, String message, RawRequest request, ChannelHandlerContext ctx) {
```
willsalz: what is this doing in the ErrorMessage? Makes no sense.

tchannel-core/src/main/java/com/uber/tchannel/handlers/MessageDefragmenter.java
```diff
@@ -60,26 +63,47 @@
...
+        if (msg instanceof CallFrame) {
+            return ((CallFrame) msg).moreFragmentsFollow();
+        }
+
+        return false;
+    }
 
+    private void decodeErrorResponse(ChannelHandlerContext ctx, ErrorFrame msg, List<Object> out) {
+
+        assert this.messageMap.get(msg.getId()) == null;
```
willsalz: this is effectively a noop

tchannel-core/src/main/java/com/uber/tchannel/schemes/ErrorResponse.java
```diff
@@ -0,0 +1,53 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.tchannel.schemes;
+
+import com.uber.tchannel.errors.ErrorType;
+import com.uber.tchannel.frames.FrameType;
+
+public final class ErrorResponse extends ResponseMessage {
```
willsalz: What is this? Where in the protocol does this exist?

tchannel-core/src/main/java/com/uber/tchannel/schemes/RawResponse.java
```diff
@@ -89,6 +96,12 @@ public ByteBuf getArg3() {
         return arg3;
     }
 
+    public void release() {
```
willsalz: this should implement ByteBufHolder if you want to treat this as a reference counted object.

willsalz:
Please remove all renaming of message -> frame. They are called `messages`. That also implies, incorrectly, that a `frame` has some correspondence with a TCP frame, which it does not.

Also, I've said this before as has @truncs, none of this peer or sub channel work should be in the core tchannel library. They are shaky abstractions, at best, that come only from the node.js implementation i.e. aren't concepts even touched on in the TChannel protocol or Hyperbahn, and add a lot of unnecessary complexity and abstraction leakage into the core lib.

tchannel-example/src/main/java/com/uber/tchannel/thrift/KeyValueClient.java
ShanniLi: Set host port in TChannel is for listen()
Here is for remote host port. This is direct connection. When "peers" are passed to subChannel, peer selection will become effective.

ShanniLi: ref: https://github.com/uber/tchannel-node/blob/master/channel.js#L835

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
ShanniLi: peers is List<InetSocketAddress>
this.peers is List<SubPeer>

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
ShanniLi: yea, TChannel supports two mode: 1) direct connection to a remote host (already supported in tchannel-java) 2) connections to a list of peers, i.e., peer selection.

The second one is most commonly used in Hyperbahn.

ShanniLi: ref: https://github.com/uber/tchannel-java/pull/89/files#diff-6f54a004122ffb4db1cbf0ed46cde280R232

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: This is fine. The only case when SubChannel needs to be GC'ed is in Hyperbahn implementation. Even in that case, it should be explicitly removed.

tchannel-core/src/main/java/com/uber/tchannel/checksum/Checksums.java
ShanniLi: They are actually frame definitions. Naming them message is confusing. I understand to have message in the name is for the sake of netty channel/pipeline. However, that should be less focused in the world of TChannel.

ShanniLi: Also it is confusing with schemes.messages

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
ShanniLi: Not sure why it has to be message. They are actually tchannel Frames and FrameTypes.

tchannel-core/src/main/java/com/uber/tchannel/frames/ErrorFrame.java
ShanniLi: It is a common utility whenever error responses need to be sent.

ref: https://github.com/uber/tchannel-java/pull/89/files#diff-0c30656be54ed88283a28294e0fa1555R94

tchannel-core/src/main/java/com/uber/tchannel/handlers/MessageDefragmenter.java
ShanniLi: ok, I will remove it.

tchannel-core/src/main/java/com/uber/tchannel/schemes/ErrorResponse.java
ShanniLi: Search error: https://github.com/uber/tchannel/blob/master/docs/protocol.md
We call it error responses/frames.
For example, when service is busy, a busy frame should be sent back to client instead of throwing an exception at the service side (the current implementation).

tchannel-core/src/main/java/com/uber/tchannel/schemes/RawResponse.java
ShanniLi: ok, I will update the code.

ShanniLi:
@willsalz OK, I will rename frames back to message.

As for peer and subChannels, it is implemented in other languages as well:
https://github.com/uber/tchannel-go/blob/master/channel.go#L109
https://github.com/uber/tchannel-go/blob/master/channel.go#L110

The protocol doc only describes the minimal requirement to talk in TChannel. It doesn't describe how to do connection pooling, retries, or how to handle failures. Even for those features you don't see in other languages yet will be added sooner or later. There is no core TChannel or Hyperbahn TChannel. It is just TChannel. Quote @jakev's words, TChannel and Hyperbahn are like Ying and Yang. 

Of course, there are Hyperbahn specific features implemented in node, which are not in other languages. E.g., relay forwarding. There are some libraries only implemented for some languages, e.g., thriftrw (node and python). What you see in my PRs is part of what necessary makes it TChannel.

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
ShanniLi: OK, I will rename them back to messages.

ShanniLi:
@willsalz We had a discussion about the naming in the protocol spec. The spec was unclear about the difference between frames and messages. @kriskowal wrote up a clarification (https://github.com/uber/tchannel/pull/1395). In the case of this implementation, it should be called frames.

anson627:
lgtm

PR #93, opened 2015-10-23
Add req/res release to fix tests

anson627:
@ShanniLi


ShanniLi:
lgtm. Need to investigate the partial build failure. But it is good for now.

PR #104, opened 2015-11-10
Better channel write logic

ShanniLi:
r @anson627 

benchmark: 70k+ qps

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
@@ -38,6 +38,7 @@ import com.uber.tchannel.messages.Request;
 import com.uber.tchannel.messages.ResponseMessage;
 import com.uber.tchannel.messages.Serializer;
 import com.uber.tchannel.messages.ThriftSerializer;
+import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
 import io.netty.util.CharsetUtil;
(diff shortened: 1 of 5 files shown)


tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -148,7 +128,15 @@ public void onSuccess(ResponseMessage response) {
 
                 // Since the request was handled, decrement the queued requests count
                 queuedRequests.decrementAndGet();
-                ctx.writeAndFlush(response);
+
+                // TODO: aggregate the flush
+                if (!ctx.channel().isWritable()) {
+                    synchronized (responseQueue) {
```
anson627: ConcurrentLinkedQueue is lock free, which might be faster?

tchannel-core/src/test/java/com/uber/tchannel/channels/ErrorResponseHandlingTest.java
```diff
@@ -86,49 +86,6 @@ public void testBadRequestErrorOnInvalidArgScheme() throws Exception {
         client.shutdown();
     }
 
-    @Test
```
anson627: why delete a test?

ShanniLi: I removed the rate-limiting, so the test won't apply anymore.

anson627:
:+1: any improvement to RPS after switching to a lock free data structure.

ShanniLi:
Actually the new version is worse. After the change, the QPS can never reach 70k again. You can try it out using PingPongServerBenchmark.

PR #105, opened 2015-11-10
Replace connection exception with response errors

ShanniLi:
r @anson627

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
@@ -235,20 +235,7 @@ public final class SubChannel {
             request.getId(),
             errorType,
             throwable));
-        switch (argScheme) {
-            case RAW:
-                ((TFuture<RawResponse>)future).set((RawResponse) response);
-                break;
-            case JSON:
-                ((TFuture<JsonResponse>)future).set((JsonResponse) response);
-                break;
-            case THRIFT:
-                ((TFuture<ThriftResponse>)future).set((ThriftResponse) response);
-                break;
-            default:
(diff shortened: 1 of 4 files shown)


anson627:
:+1:

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
```diff
@@ -225,4 +227,28 @@ protected ResponseRouter prepare(
     ) throws TChannelError {
         return send(request, null, 0);
     }
+
+    public <V> ListenableFuture<V> createError(Request request, ErrorType errorType, Throwable throwable) {
+        TFuture<V> future = TFuture.create();
+        ArgScheme argScheme = request.getArgScheme();
+        Response response = Response.build(argScheme, new ErrorResponse(
```
anson627: maybe response builder should take care of building different Response based on different ArgScheme?

tchannel-core/src/main/java/com/uber/tchannel/messages/ErrorResponse.java
```diff
@@ -30,12 +30,22 @@
...
     public ErrorResponse(long id, ErrorType errorType, String message) {
         this.id = id;
         this.errorType = errorType;
         this.message = message;
         this.type = FrameType.Error;
+        this.throwable = null;
+    }
+
+    public ErrorResponse(long id, ErrorType errorType, Throwable throwable) {
+        this.id = id;
```
anson627: you can have a private base constructor, and have both call it with different parameters.

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
ShanniLi: Refactored part of the code to ResponseRouter.

PR #106, opened 2015-11-12
Gracefully handling connection issues

ShanniLi:
r @anson627 

Several major updates:
1. gracefully handling outbound exceptions
2. gracefully handling inbound exceptions
3. reset connection on 10 continuous timeout
4. remove connection from peer when it fails
5. add benchmark for multiple connection


tchannel-core/src/test/java/com/uber/tchannel/channels/ConnectionTest.java
```diff
@@ -135,15 +202,174 @@ public void testConnectionFailure() throws Exception {
...
+            this.delayed = delayed;
+        }
 
         @Override
         public RawResponse handle(Request request) {
+
+            if (delayed) {
+                try {
+                    sleep(300);
+                } catch (Exception ex) {
```
anson627: do nothing if exception?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -153,15 +153,24 @@ public SubChannel makeSubChannel(String service) {
         return this.makeSubChannel(service, Connection.Direction.NONE);
     }
 
-    public void shutdown(boolean sync) throws InterruptedException, ExecutionException {
+    public void shutdown(boolean sync) {
```
anson627: why logging instead of exception?

anson627:
minor comments, otherwise lgtm

tchannel-core/src/test/java/com/uber/tchannel/channels/ConnectionTest.java
ShanniLi: It is InterruptedException, not too much we can do. First, it is unlikely we get interrupted here. Second, even we get interrupted, the test will fail. So, no fasle positives.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: Not sure it is a good idea to ask users to handle these two exceptions. They are shutting down tchannel, so I guess they won't care about these two exceptions at this stage.

PR #107, opened 2015-11-15
Better peer selection and max pending requests

ShanniLi:
@anson627 

Several fixes.


tchannel-benchmark/src/main/java/com/uber/tchannel/benchmarks/PingPongServerBenchmark.java
```diff
@@ -177,5 +199,14 @@ public int actualQPS() {
         public int errorQPS() {
             return errorQPS.get();
         }
+        public int timeoutQPS() {
```
anson627: why need getter for public members?

ShanniLi: This is used by the benchmark platform. It uses getter to print out the QPS result.

anson627:
minor comments, lgtm.

PR #108, opened 2015-11-18
Retry

ShanniLi:
r @anson627

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -76,7 +76,7 @@ public final class TChannel {
     private String listeningHost = "0.0.0.0";
     private int listeningPort;
     private ExecutorService exectorService;
-    private final int initTimeout;
+    private final long initTimeout;
     private final int resetOnTimeoutLimit;
     private final int clientMaxPendingRequests;
 
@@ -200,7 +200,7 @@ public final class TChannel {
         private EventLoopGroup bossGroup = new NioEventLoopGroup(1);
         private EventLoopGroup childGroup = new NioEventLoopGroup();
         private LogLevel logLevel = LogLevel.INFO;
-        private int initTimeout = 2000;
+        private long initTimeout = -1;
         private int resetOnTimeoutLimit = Integer.MAX_VALUE;
         private int clientMaxPendingRequests = 100000;
 
@@ -247,7 +247,7 @@ public final class TChannel {
             return this;
         }
 
-        public Builder setInitTimeout(int initTimeout) {
+        public Builder setInitTimeout(long initTimeout) {
             this.initTimeout = initTimeout;
             return this;
         }
--- a/tchannel-core/src/main/java/com/uber/tchannel/channels/SubPeer.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/channels/SubPeer.java
@@ -24,6 +24,7 @@ package com.uber.tchannel.channels;
 
 import com.uber.tchannel.api.SubChannel;
 import com.uber.tchannel.api.errors.TChannelError;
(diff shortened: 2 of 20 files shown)


tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -200,7 +200,7 @@ public int getClientMaxPendingRequests() {
         private EventLoopGroup bossGroup = new NioEventLoopGroup(1);
         private EventLoopGroup childGroup = new NioEventLoopGroup();
         private LogLevel logLevel = LogLevel.INFO;
-        private int initTimeout = 2000;
+        private long initTimeout = -1;
```
anson627: what does timeout = -1 mean?

tchannel-core/src/main/java/com/uber/tchannel/channels/SubPeer.java
```diff
@@ -59,7 +60,13 @@ public Peer getPeer() {
         return peerManager.getPeer(remoteAddress);
     }
 
-    public boolean updateScore() {
+    public boolean updateScore(OutRequest outRequest) {
+        if (outRequest.isUsedPeer(remoteAddress)) {
+            // if we have used this peer, it should be given a lower score
+            score = SCORE_UNCONNECTED - 0.1;
```
anson627: so peer will never be re-used?

tchannel-hyperbahn/src/test/java/com/uber/tchannel/hyperbahn/api/HyperbahnClientTest.java
```diff
@@ -63,7 +65,31 @@ public void testAdvertise() throws Exception {
...
+        }};
+
+        HyperbahnClient hyperbahnClient = new HyperbahnClient.Builder("service", tchannel)
+            .setRouters(routers)
+            .setAdvertiseInterval(100)
+            .build();
+        hyperbahnClient.advertise();
+
+        sleep(500);
+        assertTrue(responseHandler.requestsReceived >= 3);
```
anson627: why it's not == 5, since the retry limit is 4.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: If timeout == -1, it will use the req timeout:
https://github.com/uber/tchannel-java/pull/108/files#diff-6f54a004122ffb4db1cbf0ed46cde280R269

tchannel-core/src/main/java/com/uber/tchannel/channels/SubPeer.java
ShanniLi: This is for retry logic. If a peer is tried before, it will have lower score than even unconnected peers. It will be reused when all "good" peers are exhausted.

tchannel-hyperbahn/src/test/java/com/uber/tchannel/hyperbahn/api/HyperbahnClientTest.java
ShanniLi: It means one init request + 4 retries. So 5 attempts total.

ShanniLi: The is the common behavior for all languages except for node.

tchannel-core/src/main/java/com/uber/tchannel/channels/SubPeer.java
anson627: ok so we still honor retry limit. If retry limit is larger than number of peers, we will still retry by limit count.

anson627:
:+1:

PR #109, opened 2015-11-19
fix the bug in handling empty arg2

ShanniLi:
@anson627

### DIFF
--- a/tchannel-example/src/main/java/com/uber/tchannel/thrift/KeyValueClient.java
+++ b/tchannel-example/src/main/java/com/uber/tchannel/thrift/KeyValueClient.java
@@ -74,7 +74,14 @@ public class KeyValueClient {
                 8888
             );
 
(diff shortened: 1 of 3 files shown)


tchannel-example/src/main/java/com/uber/tchannel/thrift/KeyValueClient.java
```diff
@@ -104,6 +116,8 @@ public static String getValue(
             throw err;
         }
 
+        getResult.release();
```
anson627: the client always has to remember release? is there anyway to do self-release (like finalize)?

ShanniLi: yea, it is painful. I will send a separate diff to fix this.

anson627:
:+1:

PR #111, opened 2015-12-01
add utility to get current ip instead of using inetaddress.getlocalhost

ShanniLi:
r @anson627 @junchaowu


anson627:
this won't work, because getCurrentIP gets docker's ip address instead of host's.

ShanniLi:
We still need this for none docker case, right?

anson627:
lgtm

PR #112, opened 2015-12-02
Add health check

anson627:
@ShanniLi @junchaowu 

if default health check:
       registerHealthHanlder()
else if customized health check:
       subclass HealthCheckRequestHandler
       registerHealthHanlder(instance of subclass)
else:
       // no health check


ShanniLi:
lgtm

PR #113, opened 2015-12-02
Logger support

ShanniLi:
r @anson627 @junchaowu 

Use slf4j + log4j for logging.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -54,6 +54,7 @@ import org.apache.log4j.Level;
 import org.apache.log4j.LogManager;
 import org.apache.log4j.Logger;
 import org.apache.log4j.PropertyConfigurator;
+import org.slf4j.LoggerFactory;
 
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
@@ -67,6 +68,8 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 
 public final class TChannel {
+
+    private static final org.slf4j.Logger logger = LoggerFactory.getLogger(TChannel.class);
 
     private final HashedWheelTimer timer;
 
@@ -129,6 +132,10 @@ public final class TChannel {
         return this.initTimeout;
     }
 
+    public boolean hasListened() {
+        return !listeningHost.equals("0.0.0.0");
+    }
+
     public ChannelFuture listen() throws InterruptedException {
         ChannelFuture f = this.serverBootstrap.bind(this.host, this.port).sync();
         InetSocketAddress localAddress = (InetSocketAddress) f.channel().localAddress();
@@ -142,8 +149,13 @@ public final class TChannel {
         return subChannels.get(service);
     }
 
-    // TODO: make sure makeSubChannel is called before listening ...
     public SubChannel makeSubChannel(String service, Connection.Direction preferredDirection) {
+        if (hasListened()) {
+            logger.warn("makeSubChannel should be called before listen - service: {}",
+                service
+            );
+        }
+
         SubChannel subChannel = getSubChannel(service);
         if (subChannel == null) {
             subChannel = new SubChannel(service, this, preferredDirection);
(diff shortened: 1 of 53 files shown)


tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -52,10 +46,18 @@
...
 import io.netty.handler.logging.LoggingHandler;
 import io.netty.util.HashedWheelTimer;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+import io.netty.util.internal.logging.Slf4JLoggerFactory;
+import org.apache.log4j.BasicConfigurator;
+import org.apache.log4j.Level;
+import org.apache.log4j.LogManager;
+import org.apache.log4j.Logger;
+import org.apache.log4j.PropertyConfigurator;
+import org.slf4j.LoggerFactory;
```
yurishkuro: it is customary not to tie Java libraries to a particular logging implementation. Why is it not enough to just log to slf4j? The actual application then can bind to whichever logging framework best suits it.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -66,6 +68,8 @@
 
 public final class TChannel {
```
junchaowu: this is non-related. but we should remove 'final' , and allow users to override or extend TChannel class.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -134,6 +131,10 @@ public long getInitTimeout() {
         return this.initTimeout;
     }
 
+    public boolean hasListened() {
```
junchaowu: hasListened -> isListening

tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
```diff
@@ -14,15 +18,21 @@
...
 
-    public static final void PrintException(Throwable throwable) {
+    public static void setLogLevel(Level level) {
+        LogManager.getRootLogger().setLevel(level);
+    }
+
+    public static void PrintException(Throwable throwable) {
         StringWriter writer = new StringWriter();
         PrintWriter printWriter = new PrintWriter( writer );
         throwable.printStackTrace( printWriter );
```
junchaowu: why do u need printWriter? just do printStackTrace if you want to output to console.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: make sense. I will update the PR.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: Is there any reason to allow users to override or extend TChannel?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -134,6 +131,10 @@ public long getInitTimeout() {
...
+        return !listeningHost.equals("0.0.0.0");
```
anson627: I thought tchannel listens on real ip now.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -214,8 +240,13 @@ public Builder(String service) {
             this.service = service;
             this.host = TChannelUtilities.getCurrentIp();
             if (this.host == null) {
-                // TODO: logging
+                logger.error("failed to get current IP");
             }
+
+            timer = new HashedWheelTimer(10, TimeUnit.MILLISECONDS);
```
anson627: why HashedWheelTimer instead of java timer?

tchannel-core/src/main/java/com/uber/tchannel/handlers/ResponseRouter.java
```diff
@@ -138,7 +141,7 @@ protected void handleResponse(ResponseMessage response) {
         OutRequest outRequest = this.requestMap.remove(response.getId());
         if (outRequest == null) {
             response.release();
-            // TODO logging
+            // logger.warn("response without corresponding request");
```
anson627: comment out?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: if TChannel.listen() is not called, the default value for listeningHost would be "0.0.0.0"

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
ShanniLi: There can be lots of requests, java timer does not work well in such case.

anson627:
lgtm

junchaowu:
lgtm

PR #114, opened 2015-12-03
bug fixes

ShanniLi:
r @anson627 @junchaowu


anson627:
:+1:

junchaowu:
lgtm. note: the health handler should be default.

ShanniLi:
@junchaowu the health check handler should be default, since it only applies to service sub channels.

PR #115, opened 2015-12-05
Perf improvement: unblocking call, aggregated flush, and shared thread pool

ShanniLi:
r @anson627 @junchaowu


tchannel-core/src/main/java/com/uber/tchannel/messages/Response.java
```diff
@@ -179,6 +179,20 @@ public final ArgScheme getArgScheme() {
         return ArgScheme.toScheme(transportHeaders.get(TransportHeaders.ARG_SCHEME_KEY));
     }
 
+    public final long argSize() {
```
junchaowu: no need use long type. int can fit the size.

junchaowu:
lgtm

Issue #102, opened 2015-11-06
ThriftResponse implements AutoCloseable


brandoniles:
If ThriftResponse implements AutoCloseable then clients can use the try-with-resources pattern:

```
try (ThriftResponse<FooService.foo_result> response = future.get()) {
    result = response.getBody(FooService.foo_result.class);
}
```
vs
```
ThriftResponse<FooService.foo_result> response = future.get();
try {
    result = response.getBody(FooService.foo_result.class);
} finally {
    if (response != null) {
        response.release();
    }
}
```

ShanniLi:
done

PR #116, opened 2015-12-07
Improving req/res Interfaces and adding examples

ShanniLi:
r @anson627 @junchaowu 

Improving req/res interfaces based on feedbacks, e.g., #102 

Let me know how do you like the new interfaces:
1. sync'ed wait on response should use try-with-resources pattern to release the response after done.
2. async'ed response handling doesn't need to call release. TChannel library will release it.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
@@ -41,6 +41,7 @@ import com.uber.tchannel.messages.JsonResponse;
 import com.uber.tchannel.messages.RawRequest;
 import com.uber.tchannel.messages.RawResponse;
 import com.uber.tchannel.messages.Request;
+import com.uber.tchannel.messages.Response;
 import com.uber.tchannel.messages.Serializer;
 import com.uber.tchannel.messages.ThriftRequest;
 import com.uber.tchannel.messages.ThriftResponse;
@@ -61,9 +62,8 @@ public final class SubChannel {
     private final PeerManager peerManager;
     private final long initTimeout;
     private final Connection.Direction preferredDirection;
-
-    private Map<String, RequestHandler> requestHandlers = new HashMap<>();
-    private List<SubPeer> peers = new ArrayList<>();
+    private final List<SubPeer> peers = new ArrayList<>();
+    private final Map<String, RequestHandler> requestHandlers = new HashMap<>();
 
     private final Serializer serializer = new Serializer(new HashMap<ArgScheme, Serializer.SerializerInterface>() {
         {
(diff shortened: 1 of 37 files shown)


tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
```diff
@@ -226,13 +226,13 @@ public boolean sendOutRequest(OutRequest outRequest) {
         return sendRequest(request, host, port);
     }
 
-    public ListenableFuture<RawResponse> send(
+    public TFuture<RawResponse> send(
```
anson627: what's wrong with ListenableFuture? This is a breaking change?

tchannel-example/src/main/java/com/uber/tchannel/basic/AsyncRequest.java
```diff
@@ -0,0 +1,149 @@
...
+                .build();
+            TFuture<RawResponse> future = subChannel.send(request,
+                InetAddress.getByName("127.0.0.1"),
+                8888
+            );
+
+            future.addCallback(callback);
+        }
+
+        synchronized (counter) {
```
anson627: you can use CountDownLatch. But I guess it's ok here.

tchannel-core/src/main/java/com/uber/tchannel/api/TFuture.java
```diff
@@ -1,26 +1,121 @@
...
-public final class TFuture<V> extends AbstractFuture<V> {
-    public static <V> TFuture<V> create() {
-        return new TFuture();
+    private TFuture(ArgScheme argScheme) {
+        this.argScheme = argScheme;
     }
 
-    private TFuture() {
+    @SuppressWarnings({"unchecked"})
+    public void addCallback(final TFutureCallback<V> callback) {
```
anson627: not really a big fan of callback in java. The closest concept is listener, and you already have way to addListener.

junchaowu:
lgtm

tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
ShanniLi: TFuture is inherited from ListenableFuture. So, it is not breaking change. However, new implementations should use TFuture instead of ListenableFuture.

tchannel-core/src/main/java/com/uber/tchannel/api/TFuture.java
ShanniLi: The reason I added this interface is because the original interface support Futures.addCallback: 
e.g., https://github.com/uber/tchannel-java/blob/master/tchannel-hyperbahn/src/main/java/com/uber/tchannel/hyperbahn/api/HyperbahnClient.java#L102

I want to deprecate that. 

BTW, why addListener is better? The callback is implemented based on addListener with better error handlings.

anson627:
lgtm

PR #123, opened 2016-02-24
use ip address instead of hostname to initialize TChannel

nath-uber:
hyperbahn requires that the client registers with a valid IP address.

### DIFF
--- a/tchannel-core/src/test/java/com/uber/tchannel/api/TChannelBuilderTest.java
+++ b/tchannel-core/src/test/java/com/uber/tchannel/api/TChannelBuilderTest.java
@@ -47,11 +47,15 @@ public class TChannelBuilderTest extends BaseTest {
     @Test
     public void testServerListeningHostValidity() throws Exception {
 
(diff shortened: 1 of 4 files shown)


junchaowu:
can you rebase on master?

tchannel-core/src/test/java/com/uber/tchannel/api/TChannelBuilderTest.java
```diff
@@ -43,4 +44,15 @@ public void testSetServerHost() throws Exception {
...
     }
 
+    @Test
+    public void testServerListeningHostValidity() throws Exception {
+
+        TChannel tchannel = new TChannel.Builder("some-service")
+                .setServerHost(InetAddress.getByName("localhost"))
+                .build();
+        tchannel.listen();
+        System.out.println(tchannel.getListeningHost());
```
junchaowu: remove system out

junchaowu:
you may still need to rebase.

PR #125, opened 2016-03-03
Add default handler

junchaowu:
In order to integrate to yarpc-java, I need a default handle which can take whatever requests that doesn't have corresponding endpoint.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
@@ -119,6 +119,10 @@ public class RequestRouter extends SimpleChannelInboundHandler<Request> {
         // Get handler for this method
         RequestHandler handler = this.getRequestHandler(service, endpoint);
         if (handler == null) {
+            handler = topChannel.getDefaultHandler();
+        }
+
+        if (handler == null) {
             sendError(ErrorType.BadRequest,
                 "Invalid handler function",
                 request, ctx);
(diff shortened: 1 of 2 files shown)


tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -119,6 +119,10 @@ protected void messageReceived(final ChannelHandlerContext ctx, final Request re
         // Get handler for this method
         RequestHandler handler = this.getRequestHandler(service, endpoint);
         if (handler == null) {
+            handler = topChannel.getDefaultHandler();
+        }
+
+        if (handler == null) {
```
anson627: if default handler can be null, then what's the point of default handler?

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -119,6 +119,10 @@ protected void messageReceived(final ChannelHandlerContext ctx, final Request re
         // Get handler for this method
         RequestHandler handler = this.getRequestHandler(service, endpoint);
         if (handler == null) {
+            handler = topChannel.getDefaultHandler();
```
ShanniLi: why default handler can be null? You need to handle the req regardless, otherwise, there will be buffer leak

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
junchaowu: the default handler will be set from yarpc-java which is using tchannel-java. If user doesn't set the defaulthandler , the tchannel still does in the existing way. That's why the default handler can be null.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
junchaowu: You can treat this default handler as a normal user handler. It takes request and returns response. It still returns back to the requestRouter.

ShanniLi:
lgtm

PR #128, opened 2016-03-25
CodecTestUtil is not a junit test.

jwils:
Looks like it doesn't cause any errors when running the tests from mvn. My Intellij config tried to run the class as a test though and throws an error.


junchaowu:
:+1:

breerly:
Hard to complain. :shipit:

PR #129, opened 2016-03-27
Downgrade to work with netty 4.1

### LINKED ISSUE #127: Downgrade Netty version to 4.1
Netty 5.x is abandoned according to netty/netty#4466 . 
There won't be any 5.x updates or releases in the future. They have recommended users to use 4.1 version. Let's why we need to downgrade our Netty versions and fix downgrade related issues.

jwils:
Resolves #127 

Did some basic testing. Not sure how much apps depending on this are using netty 5.0 though.

Looks like there was some changes to the embedded pipelining buffers which caused a leak in the ErrorCodecTest. Other than that I there weren't any function changes needed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -224,7 +224,7 @@
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-all</artifactId>
-            <version>5.0.0.Alpha2</version>
+            <version>4.1.0.CR3</version>
         </dependency>
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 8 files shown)


junchaowu:
Right. I was stuck at the embeddedpipeline part as well. The bytebuf reference count somehow isn't working exactly same between 5.x and 4.x .

jwils:
I think before when you read from the embedded channel that you wrote to you would get the same buffer. In 4.1 I believe it makes a copy.

PR #131, opened 2016-03-30
Add support for allowing users to pass custom request router

junchaowu:
Yarpc-java would like to have own request router in order to send tchannel error directly.


ShanniLi:
test failed. Otherwise, lgtm.

PR #133, opened 2016-05-12
adjust netty version requirement

chrislusf:
This use basic netty feature so that tChannel can work with Spark.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 
     <groupId>com.uber.tchannel</groupId>
     <artifactId>tchannel</artifactId>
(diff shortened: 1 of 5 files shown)


junchaowu:
this is not going to work.

breerly:
Mind elaborating?

chrislusf:
not ready yet.

breerly:
Please remove the `WIP -` prefix from this PR when it's ready to review.

junchaowu:
We are not going to do this approach because we should change Netty's version because of Spark. It could happen again on some day when there is another library using other version's Netty, asking to change version.

PR #138, opened 2016-07-23
Borrow the getCurrentIp implementation from Go

### LINKED ISSUE #137: getCurrentIp needs consistent ordering
We're seeing a problem where tchannel applications in Java are (incorrectly) binding to eth1 instead of eth0. 

The ordering of this method varies by platform. We need to do something like `https://github.com/uber/tchannel-python/blob/master/tchannel/net.py#L44` to guarantee an order.

```java
public static InetAddress getCurrentIp() {
    try {
        Enumeration<NetworkInterface> networkInterfaces = NetworkInterface
            .getNetworkInterfaces();
        while (networkInterfaces.hasMoreElements()) {
            NetworkInterface ni = (NetworkInterface) networkInterfaces
                .nextElement();
            Enumeration<InetAddress> nias = ni.getInetAddresses();
            while(nias.hasMoreElements()) {
                InetAddress ia= (InetAddress) nias.nextElement();
                if (!ia.isLinkLocalAddress()
                    && !ia.isLoopbackAddress()
                    && ia instanceof Inet4Address) {
                    return ia;
                }
            }
        }
    } catch (SocketException e) {
        logger.error("unable to get current IP.", e);
    }
    return null;
}
```

blampe:
Resolves #137.

Java's iteration order for network interfaces looks reversed from Go. Here's Go's [`ListenIP`](https://github.com/uber/tchannel-go/blob/dev/localip.go#L59) with some logging added:
```
$ go run main.go
2016/07/23 17:15:02 Checking 127.0.0.1/8
2016/07/23 17:15:02 Checking 10.28.174.63/32
2016/07/23 17:15:02 Checking ::1/128
2016/07/23 17:15:02 Checking 10.28.40.63/24
2016/07/23 17:15:02 Checking fe80::2e60:cff:fea2:2f30/64
2016/07/23 17:15:02 Checking 10.28.104.63/24
2016/07/23 17:15:02 Checking fe80::2e60:cff:fea2:2f31/64
2016/07/23 17:15:02 Checking 172.17.0.1/16
```

And here's Java:
```
$ java IpTest
Checking 172.17.0.1
Checking fe80:0:0:0:2e60:cff:fea2:2f31%3
Checking 10.28.104.63
Checking fe80:0:0:0:2e60:cff:fea2:2f30%2
Checking 10.28.40.63
Checking 0:0:0:0:0:0:0:1%1
Checking 10.28.174.63
Checking 127.0.0.1
```

This PR mimics the Go implementation but prefers the last-seen match instead of the first-seen to account for this reverse ordering. The existing Java implementation had some other differences from Go (wasn't checking if the interface was up, for example) and so I feel more comfortable just copying Go since we know that works well.

I've confirmed locally that with this change Java will listen on the same IP as Go.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
@@ -54,7 +54,7 @@ public class TChannelUtilities {
                 }
             }
         } catch (SocketException e) {
-            logger.error("unable to score IP.", e);
+            logger.error("Unable to score IP {} of interface {}", addr, iface, e);
         }
 
         return score;
@@ -87,7 +87,7 @@ public class TChannelUtilities {
                 }
             }
         } catch (IOException e) {
-            logger.error("problem getting local IP.", e);
+            logger.error("Problem getting local IP", e);
         }
 
         return bestAddr;
--- a/tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
@@ -27,6 +27,7 @@ import io.netty.buffer.Unpooled;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
 import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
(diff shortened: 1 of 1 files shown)


tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
```diff
@@ -38,26 +39,57 @@
...
+                    if (score >= bestScore) {
+                        bestScore = score;
+                        bestAddr = addr;
                     }
                 }
             }
-        } catch (SocketException e) {
-            logger.error("unable to get current IP.", e);
+        } catch (IOException e) {
+            ;
```
nklmish: May be its better to log the exception?

blampe:
@junchaowu any pointers for how to release a new version of this?

blampe:
cc @abhinav @prashantv @nikunjgit

tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
```diff
@@ -38,26 +39,57 @@
...
+
+        try {
+            if (!iface.isLoopback() && !addr.isLoopbackAddress()) {
+                score += 100;
+                if (iface.isUp()) {
+                    score += 100;
+                }
+            }
+        } catch (SocketException e) {
+            ;
```
abhinav: log or re-raise

tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
abhinav: +1

junchaowu:
good to go. The release process will take a long process since the release people need to get authorization from the central maven. If you are in hurry, just land the diff. I will release for you.

tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
```diff
@@ -38,26 +39,57 @@
...
+
+        try {
+            if (!iface.isLoopback() && !addr.isLoopbackAddress()) {
+                score += 100;
+                if (iface.isUp()) {
+                    score += 100;
+                }
+            }
+        } catch (SocketException e) {
+            logger.error("unable to score IP.", e);
```
abhinav: Nit: Include more information

    logger.error("Unable to score IP {} of interface {}", addr, iface, e)

tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
```diff
@@ -38,26 +39,57 @@
...
+                    if (score >= bestScore) {
+                        bestScore = score;
+                        bestAddr = addr;
                     }
                 }
             }
-        } catch (SocketException e) {
-            logger.error("unable to get current IP.", e);
+        } catch (IOException e) {
+            logger.error("problem getting local IP.", e);
```
abhinav: Nit: "Problem getting local IP"

abhinav:
lgtm with nits

abhinav:
:shipit:

tchannel-core/src/main/java/com/uber/tchannel/utils/TChannelUtilities.java
```diff
@@ -38,26 +39,57 @@
...
 
+    public static int scoreAddr(NetworkInterface iface, InetAddress addr) {
+        int score = 0;
+
+        if (addr instanceof Inet4Address) {
+            score += 300;
+        }
+
+        try {
+            if (!iface.isLoopback() && !addr.isLoopbackAddress()) {
```
venkat1109: 1) Don't you want to also filter out or down score the link local addresses i.e. 169.x.x.x ? 
2) And give preference for 10.x.x.x over 172.x.x.x  ?

And what's the expected behavior when we have both eth0/eth1 (which is very likely) ?

malandrew: We prefer eth0 when both are available.

blampe: @venkat1109 check my example in the description, this changes it to prefer 10 over 179 and (less obviously) to prefer eth0 over eth1. Link-local addresses are scored lower.

PR #142, opened 2016-08-15
Integrate with OpenTracing API

yurishkuro:
* [x] main tracing logic
 * [x] basic propagation test
 * [x] nested propagation test for JSON
 * [x] nested propagation test for mixed encodings
 * [x] apply licences
 * [x] clean-up TODOs
 * [x] crossdock tests - separate branch https://github.com/uber/tchannel-java/tree/crossdock-tests
 * [x] use official jaeger release before merging


yurishkuro:
@blampe @abhinav ready for review

blampe:
cc @junchaowu as an fyi

junchaowu:
Overall look good.

tchannel-core/src/main/java/com/uber/tchannel/messages/EncodedRequest.java
```diff
@@ -69,6 +70,14 @@ public String getHeader(String key) {
         return getHeaders().get(key);
     }
 
+    public void updateHeaders(Map<String, String> headers) {
```
abhinav: Nit: I would call this setHeaders to make it clear that existing headers won't be retained

tchannel-core/src/main/java/com/uber/tchannel/messages/JsonRequest.java
```diff
@@ -65,8 +65,8 @@ public Builder(String service, ByteBuf arg1) {
         }
 
         @Override
-        public Builder<T> setTimeout(long timeout) {
-            super.setTimeout(timeout);
+        public Builder<T> setTimeout(long timeoutMillis) {
+            super.setTimeout(timeoutMillis);
```
abhinav: thanks!

tchannel-core/src/main/java/com/uber/tchannel/tracing/CachedKeyTransformer.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+package com.uber.tchannel.tracing;
+
+import com.google.common.base.Function;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+class CachedKeyTransformer implements Function<String, String> {
```
abhinav: What's the reason for implementing this manually rather than using `CacheBuilder` with `CacheLoader.from(function)`?

tchannel-core/src/main/java/com/uber/tchannel/tracing/PrefixedHeadersCarrier.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Map;
+
+class PrefixedHeadersCarrier implements TextMap {
+    /**
+     * Prefix used with all keys generated by the tracer when marshalling SpanContext
+     * to a map of headers.
+     * <p>
+     * This string value is shared across all implementations of TChannel.
+     */
+    private static final String TRACING_KEY_PREFIX = "$tracing$";
```
abhinav: I would make this public somewhere (not necessarily in this class) because this exact key is sort-of part of the API now.

tchannel-core/src/main/java/com/uber/tchannel/tracing/PrefixedHeadersCarrier.java
```diff
@@ -0,0 +1,124 @@
...
+                    }
+                }),
+                new Function<Map.Entry<String, String>, Map.Entry<String, String>>() {
+                    @Override
+                    public Map.Entry<String, String> apply(Map.Entry<String, String> entry) {
+                        return Maps.immutableEntry(
+                                decoder.apply(entry.getKey()),
+                                entry.getValue());
+                    }
+                });
```
abhinav: I like Guava's functional primitives as much as the next person, but I think
this code would be more readable without the `Predicate` and `Function` noise.
Even the authors of Guava have something similar to say on this: https://github.com/google/guava/wiki/FunctionalExplained#caveats

Consider using one of the following,

- Create an `AbstractIterator` (provided by Guava) inline:

    ```python
    final Iterator<Map.Entry<String, String>> headers = ...;
    
    return new AbstractIterator<Map.Entry<String, String>>() {
        protected Map.Entry<String, String> computeNext() {
            while (in.hasNext()) {
                Map.Entry<String, String> entry = in.next();
                if (!entry.getKey().startsWith(prefix)) {
                    continue;
                }
                
                return Maps.immutableEntry(
                    decoder.apply(entry.getKey()),
                    entry.getValue(),
                );
            }
            return this.endOfData();
        }
    }
    ```

- Create a private inner class.
    
    ```java
    @Override
    public Iterator<Map.Entry<String, String>> iterator() {
        return new PrefixIterator(headers.entrySet().iterator());
    }
    
    private class PrefixIterator extends AbstractIterator<...> {
        // ...
    }
    ```

- Move the construction of the `Function` and `Predicate` into helper
  functions.

    ```java
    @Override
    public Iterator<Map.Entry<String, String>> iterator() {
        return Iterators.transform(
            Iterators.filter(headers.entrySet.iterator(), prefixMatcher),
            keyDecoder,
        );
    }
    ```

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
```diff
@@ -0,0 +1,129 @@
...
+                }
+                span.finish();
+            }
+        });
+    }
+
+    public static Span startInboundSpan(
+            Request request,
+            Tracer tracer,
+            TracingContext tracingContext) {
```
abhinav: nitty nit: move `) {` to the next line

tchannel-core/src/test/java/com/uber/tchannel/tracing/TracingContextTest.java
```diff
@@ -0,0 +1,101 @@
...
+import com.uber.jaeger.samplers.Sampler;
+import com.uber.tchannel.BaseTest;
+import io.opentracing.Span;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.EmptyStackException;
+
+import static org.junit.Assert.*;
```
abhinav: nit: consider configuring your IDE to never `import *`. For IntelliJ you just have to tell it to `import *` for an unreasonably large number.

abhinav:
LGTM besides styling/readability comments.

abhinav:
:shipit:

Issue #72, opened 2015-09-20
Return ListenableFuture in the handler instead of the actual Response message


truncs:
This would make easier to write proxy servers using tchannel-java. ListenableFuture. Its a little tricky since ListenableFutures are not settable and  Guava provides another future called as SettableFuture.
So we need to compose a Future which can be Settable or Listenable to a ListenableFuture so that we can attach OnSuccess, onFailure handlers to it.

willsalz:
Super pumped for this.

manasdk:
I plan to push a PR that will introduce a new handler that returns a `ListenableFuture<? extends Response>` and call the new type a ThriftAsyncRequestHandler. This should essentially resolve the issue.

blampe:
@manasdk that's awesome! Please include @abhinav and @vivekrsharma on the PR.

PR #146, opened 2016-10-20
ThriftAsyncFutureHandler and AsyncRequestHandler

manasdk:
- Both types are meant to support async handling of requests. AsyncRequestHandler provides
  a way for handler to return a future rather than a synchonous response like the
  RequestHandler.
- ThriftAsyncRequestHandler further scopes it down to Thrift handlers and allowing those to
  return futures

This should resolve https://github.com/uber/tchannel-java/issues/72

 @abhinav and @vivekrsharma you guys might be interested.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
@@ -22,6 +22,7 @@
 
 package com.uber.tchannel.api.handlers;
 
+import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.uber.tchannel.messages.Request;
 import com.uber.tchannel.messages.Response;
@@ -40,8 +41,8 @@ public abstract class ThriftAsyncRequestHandler<T, U> implements AsyncRequestHan
         try {
             return this.handleAsync(request).get();
         } catch (InterruptedException | ExecutionException ex) {
-            // Cannot do much better with this exception so re-raising a RuntimeException.
-            throw new RuntimeException(ex);
+            // Cannot do much better with this exception so propogate the exception.
+            throw Throwables.propagate(ex);
         }
     }
 
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2015 Uber Technologies, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package com.uber.tchannel.api.handlers;
+
+import com.google.common.util.concurrent.ListenableFuture;
+import com.uber.tchannel.messages.Request;
+import com.uber.tchannel.messages.Response;
+import com.uber.tchannel.messages.ThriftRequest;
+import com.uber.tchannel.messages.ThriftResponse;
+
+import java.util.concurrent.ExecutionException;
+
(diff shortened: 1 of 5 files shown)


tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
```diff
@@ -0,0 +1,63 @@
...
+        } catch (InterruptedException | ExecutionException ex) {
+            // Cannot do much better with this exception so re-raising a RuntimeException.
+            throw new RuntimeException(ex);
+        }
+    }
+
+    @Override
+    @SuppressWarnings({"unchecked"})
+    public ListenableFuture<? extends Response> handleAsync(Request request) {
+        return handleImpl((ThriftRequest<T>)request);
```
vprithvi: This will break tracing because this does not create a new span when handling this request. 
Wrapping the tracing logic around the request like how [CallableHandler](https://github.com/uber/tchannel-java/blob/5f18d0645750edf84168b3974bf8f804fb923039/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java#L226) does it would be sufficient. 

For an overview on tracing, I suggest reading http://opentracing.io/documentation/

tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
```diff
@@ -0,0 +1,63 @@
...
+ * to be handle asynchronously by letting the actual handler return a future.
+ */
+public abstract class ThriftAsyncRequestHandler<T, U> implements AsyncRequestHandler {
+    @Override
+    public Response handle(Request request) {
+        try {
+            return this.handleAsync(request).get();
+        } catch (InterruptedException | ExecutionException ex) {
+            // Cannot do much better with this exception so re-raising a RuntimeException.
+            throw new RuntimeException(ex);
```
abhinav: Since you're already using Guava, consider using [`throw Throwables.propagate(ex)`](https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Throwables.html#propagate(java.lang.Throwable)) here instead.

tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
manasdk: Fixed tracing.

tchannel-core/src/main/java/com/uber/tchannel/api/handlers/ThriftAsyncRequestHandler.java
manasdk: Fixed.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -206,6 +216,37 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
         }
     }
 
+    private ListenableFuture<Response> sendRequestToAsyncHandler(
+            final AsyncRequestHandler asyncHandler, final Request request) {
+        // span used to trace this request
+        final Span span = Tracing.startInboundSpan(request,
+            topChannel.getTracer(), topChannel.getTracingContext());
```
vprithvi: The `Tracer` and `TracingContext` are only present when the channel is created with them, so you'll have to check whether the `Tracer` is null and fallback to the previous behavior.

tchannel-core/src/test/java/com/uber/tchannel/handlers/TestThriftAsyncRequestHandler.java
```diff
@@ -0,0 +1,109 @@
...
+
+public class TestThriftAsyncRequestHandler extends BaseTest {
+
+    private static Tracer tracer;
+    private static InMemoryReporter reporter;
+    private static TracingContext tracingContext;
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        reporter = new InMemoryReporter();
```
vprithvi: This test doesn't have any assertions on the tracing functionality. We have a  [TracingPropagationTest](https://github.com/uber/tchannel-java/blob/5f18d0645750edf84168b3974bf8f804fb923039/tchannel-core/src/test/java/com/uber/tchannel/tracing/TracingPropagationTest.java) which would be a better place to add tests for tracing. Then, this test can simply verify that the handler behaves as expected.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
yurishkuro: @vprithvi you are correct. The way it was handled in in the earlier diff is here: https://github.com/uber/tchannel-java/commit/824cf0937f920e7e6cc6e74118f1c5d98c315b89#diff-0c30656be54ed88283a28294e0fa1555R223

There has been a recent commit in opentracing-java that extracted Noop tracer/span into a reusable module. We could refactor tracing logic to return that noop span when tchnanel is not initialized with it, and not propagate that span at all, only use it as local variable here.  But that's better done as a separate PR. Here the best (although dirty) way is to check for span==null, with a TODO to fix later.

yurishkuro:
LGTM

vprithvi:
@manasdk I think the comment about the test is unaddressed. Everything else looks great!

tchannel-core/src/test/java/com/uber/tchannel/handlers/TestThriftAsyncRequestHandler.java
manasdk: Working on this now. Should have this out soon.

manasdk:
@vprithvi Done. I have finally addressed the test issues as well. Thanks for you help offline in getting this resolved.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
manasdk: Fixed.

tchannel-core/src/test/java/com/uber/tchannel/handlers/TestThriftAsyncRequestHandler.java
manasdk: Fixed

manasdk:
@abhinav Any idea why Travis fails? Does not look related to any code in this PR.

manasdk:
If travis passes on https://github.com/uber/tchannel-java/pull/147 will then merge this one at that point.

PR #153, opened 2017-01-23
Fix #152 AsyncRequestHandler callback clears TracingContext prematurely

### LINKED ISSUE #152: AsyncRequestHandler callback clears TracingContext prematurely
When `RequestRouter` sends the request to `AsyncRequestHandler` it adds a callback to finish tracing `Span` and clear `TracingContext`.
The callback doesn't get an executor to run on hence gets executed in the context of the thread that resolved the `Response` future, nuking the `TracingContext` in the middle of a thread-local span, causing an `EmptyStackException` (when the outer code tries to pop it's span off the stack), causing thread termination and other ugliness.

yborovikov:
Fixes #152.
@manasdk @abhinav @vivekrsharma

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
@@ -39,6 +39,7 @@ import com.uber.tchannel.messages.Request;
 import com.uber.tchannel.messages.Response;
 import com.uber.tchannel.messages.ResponseMessage;
 import com.uber.tchannel.tracing.Tracing;
+import com.uber.tchannel.tracing.TracingContext;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.SimpleChannelInboundHandler;
@@ -229,8 +230,7 @@ public class RequestRouter extends SimpleChannelInboundHandler<Request> {
             span = null;
         }
         final SettableFuture<Response> responseFuture = SettableFuture.create();
-        ListenableFuture<Response> handlerResponseFuture =
-            (ListenableFuture<Response>) asyncHandler.handleAsync(request);
+        ListenableFuture<? extends Response> handlerResponseFuture = asyncHandler.handleAsync(request);
         // Add callback handlers that close out the tracing span and then proxy the response.
         Futures.addCallback(handlerResponseFuture, new FutureCallback<Response>() {
             @Override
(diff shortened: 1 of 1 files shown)


manasdk:
+1

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -251,11 +251,14 @@ public void onFailure(Throwable e) {
             private void doRequestEndProcessing() {
                 if (span != null) {
                     span.finish();
+                    TracingContext tracingContext = topChannel.getTracingContext();
```
yurishkuro: imo any use of tracing context at this point will cause errors, because there is no guarantee that this Future callback will run on the same thread as line 227 that pushed the span to the context. That is why I suggest instead to add a call to `topChannel.getTracingContext().clear()` just before line 227, and in the callback only call `span.finish()`, without messing around with the context.

yborovikov: that's the point of using instrumented executor that propagates the tracing context, isn't it?
of course, it's set up outside of this code, so responsibility is on those who want to use specific tracing implementation to properly instrument the execution environment.

`.sendRequestToAsyncHandler()` that initiates the tracing is executed on the same thread as `.handleAsync()` does, so context gets full chance to be properly propagated.
the thread that resolves the response future should also be instrumented to have the context, so propagation is covered on the way back, too.

yurishkuro: Do you mean that `listeningExecutorService` is instrumented to propagate tracing context? Where is it initialized?

yborovikov: it would be initialized via `TChannel.Builder.setExecutorService()`. the latter would be called by `XTChannel.setupTChannel()` [TODO] or by whoever sets up the `TChannel` instance with thread-local-bound trace implementation.

yurishkuro: ok, fair enough. I think it would be useful to have a unit test that proves that this whole async use case works.

yborovikov: thank you for the review and your insight!
we also feel like thread-local context propagation in async flows is rather fragile and requires proper setup and participation from all moving parts.
going to merge the pr and publish new artifacts.

PR #154, opened 2017-01-26
Fix #152 empty trace context stack in async flow

### LINKED ISSUE #152: AsyncRequestHandler callback clears TracingContext prematurely

yborovikov:
Fix for `EmptyStackException` on response post-processing in `RequestRouter.sendRequestToAsyncHandler()` callback.


manasdk:
+1

PR #160, opened 2017-02-23
Reactor improvements

alshopov:
Start of improvements on build system

I intend to:
eliminate maven-antrun-plugin in preference to maven thrift plugin.
change code to enable updates of dependencies

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
@@ -43,10 +43,10 @@ import io.netty.channel.ChannelHandlerContext;
  * their current status, e.g., connected, identified, etc.
  */
 public class Peer {
-    public ConcurrentHashMap<ChannelId, Connection> connections = new ConcurrentHashMap<>();
-    public SocketAddress remoteAddress = null;
+    public final ConcurrentHashMap<ChannelId, Connection> connections = new ConcurrentHashMap<>();
+    public final SocketAddress remoteAddress;
 
-    private PeerManager manager;
+    private final PeerManager manager;
 
     public Peer(PeerManager manager, SocketAddress remoteAddress) {
         this.manager = manager;
@@ -126,9 +126,7 @@ public class Peer {
 
     public Connection getConnection(ConnectionState preferedState, Connection.Direction preferredDirection) {
         Connection conn = null;
-        Connection next = null;
-        for (ChannelId id : connections.keySet()) {
-            next = connections.get(id);
+        for (Connection next : connections.values()) {
             if (next.satisfy(preferedState)) {
                 conn = next;
                 if (preferredDirection == Connection.Direction.NONE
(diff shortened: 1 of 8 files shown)


alshopov:
Hmm, will have to look at the build failure

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -152,31 +150,25 @@ public Connection getConnection(ChannelId channelId) {
...
+        for (Connection conn : connections.values()) {
+            if (conn.direction == Connection.Direction.OUT) {
                 out++;
             } else {
                 in++;
             }
         }
 
-        Map<String, Integer> result = new HashMap<>();
+        Map<String, Integer> result = new HashMap<>(2);
```
yborovikov: Unfortunately, passing `2` to the constructor won't allow the map to hold 2 entries without re-sizing (have to account for the load factor, which is `0.75` by default). Has to be something like:
```
public static int hashMapCapacity(int size) { return (size * 4 + 2) / 3; } // for load factor of 0.75
...
   new HashMap(hashMapCapacity(2), 0.75f);
```

alshopov: You are right about the 2 and resize but the reason is different: while 2 is enough, there is an additional internal HashMap call to resize when the capacity has been reached. I am making it 3. The load factor is not a problem.

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -152,31 +150,25 @@ public Connection getConnection(ChannelId channelId) {
     }
 
     public void close() {
-        for (ChannelId id : connections.keySet()) {
-            Connection conn = connections.get(id);
-            if (conn != null) {
-                conn.close();
-            }
+        for ( Connection conn : connections.values()) {
```
manasdk: nit : space betweeen `( Connection`

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -152,31 +150,25 @@ public Connection getConnection(ChannelId channelId) {
...
+            conn.close();
```
manasdk: null check not required any longer?

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
```diff
@@ -152,31 +150,25 @@ public Connection getConnection(ChannelId channelId) {
...
     public Map<String, Integer> getStats() {
         int in = 0;
         int out = 0;
-        for (ChannelId id : connections.keySet()) {
-            Connection conn = connections.get(id);
-            if (conn == null) {
-                continue;
-            } else if (conn.direction == Connection.Direction.OUT) {
+        for (Connection conn : connections.values()) {
+            if (conn.direction == Connection.Direction.OUT) {
```
manasdk: was removal of null check intentional?

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
alshopov: Yes. connections is ConcurrentHashMap. It cannot be null. 
Like Hashtable but unlike HashMap, this class does not allow null to be used as a key or value.
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html

tchannel-core/src/main/java/com/uber/tchannel/channels/Peer.java
alshopov: Yes. connections is ConcurrentHashMap. It cannot be null. 
Like Hashtable but unlike HashMap, this class does not allow null to be used as a key or value.
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentHashMap.html

PR #162, opened 2017-03-08
Use native epoll transport if available

yborovikov:
With this patch `TChannel` will use [native epoll transport](http://netty.io/wiki/native-transports.html) if it's available.


yborovikov:
Fixed by 7c38efe.

yborovikov:
already merged.

PR #164, opened 2017-03-21
Updates with code changes

alshopov:
These updates of dependencies need code changes.
Every module update is in own commit.
The Netty update showed a problem with some of the unit tests that depended a bit too much on the internals of Netty - so first changed the unit tests to be more resilient and then I updated the dependency so even middle commit states should be fine.


abhinav:
LGTM

PR #165, opened 2017-03-22
Remove maven-antrun-plugin and start using maven-thrift-plugin

alshopov:


yborovikov:
This is an old PR, some fixes were already applied, others probably won't easily merge.

PR #175, opened 2017-09-16
Fix potential leaks in RequestRouter

yborovikov:
Fix potential leaks in `RequestRouter`;
fix a leak in `HyperbahnClientTest`;
fix a leak in `HyperbahnExample`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -165,6 +165,11 @@
                 <version>0.20.6</version>
             </dependency>
             <dependency>
+                <groupId>org.jetbrains</groupId>
+                <artifactId>annotations</artifactId>
+                <version>15.0</version>
+            </dependency>
+            <dependency>
                 <groupId>org.openjdk.jmh</groupId>
                 <artifactId>jmh-core</artifactId>
                 <version>${jmh.version}</version>
--- a/tchannel-core/pom.xml
+++ b/tchannel-core/pom.xml
@@ -62,6 +62,11 @@
             <artifactId>libthrift</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -84,7 +84,6 @@ public final class TChannel {
     private final int port;
     private String listeningHost = "0.0.0.0";
     private int listeningPort;
-    private ExecutorService exectorService;
     private final long initTimeout;
     private final int resetOnTimeoutLimit;
     private final int clientMaxPendingRequests;
@@ -97,7 +96,6 @@ public final class TChannel {
 
     private TChannel(Builder builder) {
         this.service = builder.service;
-        this.exectorService = builder.executorService;
         this.serverBootstrap = builder.serverBootstrap(this);
         this.bossGroup = builder.bossGroup;
         this.childGroup = builder.childGroup;
@@ -247,7 +245,7 @@ public final class TChannel {
         private int clientMaxPendingRequests = 100000;
         private static final int WRITE_BUFFER_LOW_WATER_MARK = 8 * 1024;
         private static final int WRITE_BUFFER_HIGH_WATER_MARK = 32 * 1024;
-        
+
         private Tracer tracer;
         private TracingContext tracingContext = new TracingContext.Default();
 
@@ -329,7 +327,7 @@ public final class TChannel {
                 .channel(useEpoll ? EpollSocketChannel.class : NioSocketChannel.class)
                 .handler(this.channelInitializer(false, topChannel))
                 .option(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT)
-                .option(ChannelOption.WRITE_BUFFER_WATER_MARK, 
+                .option(ChannelOption.WRITE_BUFFER_WATER_MARK,
                         new WriteBufferWaterMark(WRITE_BUFFER_LOW_WATER_MARK, WRITE_BUFFER_HIGH_WATER_MARK))
                 .validate();
         }
@@ -343,7 +341,7 @@ public final class TChannel {
                 .childHandler(this.channelInitializer(true, topChannel))
                 .childOption(ChannelOption.SO_KEEPALIVE, true)
                 .childOption(ChannelOption.ALLOCATOR, PooledByteBufAllocator.DEFAULT)
-                .option(ChannelOption.WRITE_BUFFER_WATER_MARK, 
+                .option(ChannelOption.WRITE_BUFFER_WATER_MARK,
                         new WriteBufferWaterMark(WRITE_BUFFER_LOW_WATER_MARK, WRITE_BUFFER_HIGH_WATER_MARK))
                 .validate();
         }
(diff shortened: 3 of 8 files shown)


tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
```diff
@@ -69,12 +69,15 @@ public static void main(String[] args) throws Exception {
         future.addCallback(new TFutureCallback<JsonResponse<AdvertiseResponse>>() {
             @Override
             public void onResponse(JsonResponse<AdvertiseResponse> response) {
-                if (!response.isError()) {
-                    System.out.println("Got response. All set: " + response.getBody(AdvertiseResponse.class));
-                } else {
-                    System.out.println("Error happened: " + response.getError().getMessage());
+                try {
+                    if (!response.isError()) {
```
alshopov: I believe isError should not throw. APIs where checking for errors can throw an error are very bad design

pom.xml
```diff
@@ -165,6 +165,11 @@
                 <version>0.20.6</version>
             </dependency>
             <dependency>
+                <groupId>org.jetbrains</groupId>
+                <artifactId>annotations</artifactId>
+                <version>15.0</version>
```
alshopov: Can yoou set "provided" here and have it propagated?

tchannel-core/pom.xml
```diff
@@ -62,6 +62,11 @@
             <artifactId>libthrift</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <scope>provided</scope>
```
alshopov: Cab provided be inherited?

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -84,7 +84,6 @@
     private final int port;
     private String listeningHost = "0.0.0.0";
     private int listeningPort;
-    private ExecutorService exectorService;
```
alshopov: I am not sure that this has to go, but since it is not used having it private is not wise. Perhaps protected.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -87,37 +89,27 @@ protected void channelRead0(final ChannelHandlerContext ctx, final Request reque
...
-        final ArgScheme argScheme = ArgScheme.toScheme(
-                request.getTransportHeaders().get(TransportHeaders.ARG_SCHEME_KEY)
-        );
-
-        if (argScheme == null) {
-            sendError(ErrorType.BadRequest,
-                "Expected incoming call to have \"as\" header set",
-                request, ctx);
+        if (request.getArgScheme() == null) {
+            sendError(ErrorType.BadRequest, "Expected incoming call to have \"as\" header set", request, ctx);
```
alshopov: Expected incoming call to have the \"as\" header set - perhaps definite article??? not sure

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -87,37 +89,27 @@ protected void channelRead0(final ChannelHandlerContext ctx, final Request reque
...
-                request, ctx);
+            sendError(ErrorType.BadRequest, "Expected incoming call to have serviceName", request, ctx);
             return;
         }
 
         // Get the endpoint. The assumption over here is that endpoints are
         // always going to to utf-8 encoded.
-        String endpoint = request.getArg1().toString(CharsetUtil.UTF_8);
+        String endpoint = request.getEndpoint();
         if (endpoint == null || endpoint.isEmpty()) {
```
alshopov: why both checks - null & isEmpty? What is the difference between these? There should be precondition on constructing the request or its setters

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -126,47 +118,49 @@ protected void channelRead0(final ChannelHandlerContext ctx, final Request reque
...
-            AsyncRequestHandler asyncHandler = (AsyncRequestHandler) handler;
-            responseFuture = sendRequestToAsyncHandler(asyncHandler, request);
-        } else {
-            // Handle the request in a separate thread and get a future to it
-            responseFuture = listeningExecutorService.submit(
-                new CallableHandler(handler, topChannel, request));
+        ListenableFuture<? extends Response> responseFuture;
+        try {
+            // In case of an AsyncRequestHandler there's no need to submit a task on the executor.
+            // It does require a down-cast to AsyncRequestHandler.
```
alshopov: This whole block seems weird design (I realise you only mostly reformatted and added the try block

pom.xml
yborovikov: i used to do that but it violates the principle of least surprise:
someone adding the dependency as
```
<dependency>
  <groupId>...</groupId>
  <artifactId>...</artifactId>
</dependency>
```
and expecting it to get (the default) `<scope>compile</scope>` would get the overridden scope (`provided`, `test`, etc.) instead.

tchannel-core/pom.xml
yborovikov: specifying it here is more obvious/fail-safe.

tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
yborovikov: this member was never used - while creating an impression it is somehow related to how the class works.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
yborovikov: yeah, we could go over error messages if there is a need - in a separate diff, perhaps.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
yborovikov: yes, there should be - but given the lack of nullability contracts in the code it's ok to leave the check in place.
checking for `null` before invoking a method on the object provides protection from `NullPointerException` in the `RequestRouter` code.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
yborovikov: that's an artifact of dealing with essentially two separate interfaces: original `RequestHandler` and `AsyncRequestHandler` that was added later.

tchannel-example/src/main/java/com/uber/tchannel/hyperbahn/HyperbahnExample.java
yborovikov: the `isError()` won't (at least shouldn't) throw but `System.out` (or whatever actual code a client can use here) could. hence the `try { ... } finally { response.release(); }` here - as an example of how the `response` should be released no matter what.

alshopov: Oh, perhaps add a comment about this

yborovikov:
@ashopov, thank you for reviewing the changes! i'm going to merge the PR in about half an hour (unless there's other feedback to address).

pom.xml
alshopov: I see. Whoever does use the version should check their effective pom in any case.
If this pattern is used throughout - then it is fine.

tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
alshopov: I understand the NPE check, what is the difference with isEmpty then? I get you point that the code as is should check it.

yborovikov: well, `.isEmpty()` checks if the string is empty, i.e. its length is `0`.

PR #178, opened 2017-10-03
Add request span interceptors

yborovikov:
This change allows to decorate a `TracingContext` implementation with the new `RequestSpanInterceptor` interface, providing an ability to intercept incoming and/or outgoing requests right after their tracing `Span`s have been created (and populated with their baggage).

The decorated implementation can inspect the `Request` and corresponding `Span` and validate these and/or tweak the span baggage.

If the interceptor method throws an exception the request is considered to fail immediately.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/tracing/RequestSpanInterceptor.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/tracing/RequestSpanInterceptor.java
@@ -5,14 +5,14 @@ import io.opentracing.Span;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * A {@link TracingContext} implementation decorated with this interface will get an option to intercept the inbound or
- * outgoing request right after the tracing {@link Span} has been created.
+ * A {@link TracingContext} implementation decorated with this interface will get an option to intercept an inbound
+ * and/or outgoing request right after the tracing {@link Span} has been created and populated.
  */
 public interface RequestSpanInterceptor {
 
     /**
-     * This method will be invoked right after the inbound request span has been created, allowing the implementation
-     * to inspect the request and/or additional baggage from the span context.
+     * This method will be invoked right after the inbound request span has been created, allowing the implementation to
+     * inspect the request and/or additional baggage from the span context.
      *
      * @throws RuntimeException
      *     if the inbound request should fail immediately
--- a/tchannel-core/src/main/java/com/uber/tchannel/tracing/RequestSpanInterceptor.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/tracing/RequestSpanInterceptor.java
@@ -1,0 +1,31 @@
+package com.uber.tchannel.tracing;
+
+import com.uber.tchannel.messages.Request;
+import io.opentracing.Span;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A {@link TracingContext} implementation decorated with this interface will get an option to intercept the inbound or
+ * outgoing request right after the tracing {@link Span} has been created.
+ */
+public interface RequestSpanInterceptor {
+
+    /**
+     * This method will be invoked right after the inbound request span has been created, allowing the implementation
+     * to inspect the request and/or additional baggage from the span context.
+     *
+     * @throws RuntimeException
+     *     if the inbound request should fail immediately
+     */
+    void interceptInbound(@NotNull Request request, @NotNull Span span) throws RuntimeException;
+
+    /**
+     * This method will be invoked right after the outgoing request span has been created, allowing the implementation
+     * to inspect the request and/or attach additional baggage to the span context.
+     *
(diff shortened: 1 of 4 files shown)


tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
```diff
@@ -31,6 +32,8 @@
 import io.opentracing.Tracer;
 import io.opentracing.propagation.Format;
 import io.opentracing.tag.Tags;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
```
yurishkuro: is this a sensible dependency?

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
```diff
@@ -86,7 +104,7 @@ public static void startOutboundSpan(
                 logger.error("Failed to inject span context into headers", e);
             }
         }
-        outRequest.getFuture().addCallback(new TFutureCallback<Response>() {
+        outRequest.getFuture().addCallback(new TFutureCallback<V>() {
```
yurishkuro: why `<V>`? It's not used.

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
```diff
@@ -50,28 +53,43 @@
...
-            Tracer tracer,
-            TracingContext tracingContext
-    ) {
-        if (tracer == null) {
+    /**
+     * @throws RuntimeException
+     *     if the outbound request should fail immediately
+     */
+    public static <V extends Response> void startOutboundSpan(
+            @NotNull OutRequest<V> outRequest,
```
yurishkuro: I don't think you need `<V extends Response>` in the method, just `OutRequest<? extends Response> outRequest`, since you don't use `V` in any way

tchannel-core/src/main/java/com/uber/tchannel/tracing/RequestSpanInterceptor.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.uber.tchannel.messages.Request;
+import io.opentracing.Span;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A {@link TracingContext} implementation decorated with this interface will get an option to intercept an inbound
+ * and/or outgoing request right after the tracing {@link Span} has been created and populated.
+ */
+public interface RequestSpanInterceptor {
```
yurishkuro: you may want to split this into InboundInterceptor and OutboundInterceptor. Otherwise it enforces unnecessary coupling on the implementation.

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
yborovikov: We use these in other places, too. The benefit is the clear contract for the arguments / return values.

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
yborovikov: Without the type parameter, the code (in `SubChannel`) has to resort to raw types. I'll update the calling site in another diff, as a part of the larger clean-up effort.

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
yborovikov: With the change in line 60 `outRequest.getFuture()` is now parameterized with the `<V>`, hence the callback has to be properly parameterized, too.

tchannel-core/src/main/java/com/uber/tchannel/tracing/RequestSpanInterceptor.java
yborovikov: I contemplated the dilemma and couldn't decide either way, so I used the simpler approach with fewer interfaces / moving parts. If you have a strong opinion/arguments towards the separate interfaces - I'll be happy to switch. The only use case I have in mind is utilizing both methods.

yborovikov:
> This look like a reasonable approach. I personally think it would've been cleaner & more generic if instead of requiring the user to extend the tracingContext with the interceptor interface we did this:

Thank you for the thorough review! I fully agree the more generic approach would be cleaner, and that it would require more design and thought to analyze all possible applications for the interceptors (and their lifecycle and interoperability contract).

> I suggest adding unit tests for the new extension.

Yes, these are coming up. Wanted to get the initial feedback on the approach as a whole, first.

yborovikov:
Ok, I've added the tests (bringing the diff coverage to 92.59% :) - unless there are any objections I'm going to merge this PR in a few minutes. Thank you for reviewing it!

PR #180, opened 2017-11-09
Fix potential response leak on RuntimeException/Error in TFuture callback

yborovikov:
Examples, benchmark tests, and `HyperbahnClient` adjusted to rely on `TFuture` callback handling to release the response.


manasdk:
👍

PR #183, opened 2017-11-14
Concurrency fixes

alshopov:
This batch is trickier.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
@@ -40,6 +40,7 @@ import java.util.Map;
 public class Connection {
 
     private static final Logger logger = LoggerFactory.getLogger(Connection.class);
+    private static final String EPHEMERAL = "0.0.0.0:0";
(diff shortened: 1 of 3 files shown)


tchannel-core/src/main/java/com/uber/tchannel/channels/Connection.java
```diff
@@ -56,64 +58,77 @@ public Connection(Peer peer, @NotNull Channel channel, Direction direction) {
         if (channel.isActive() && this.state == ConnectionState.UNCONNECTED) {
             this.state = ConnectionState.CONNECTED;
         }
+        lock = new Object();
```
yborovikov: Any reason to introduce an explicit object for locking - instead of locking on the instance itself?

alshopov: Connection is meant to be public. If someone syncs on it - it will break. If class extends it and syncs on this - it will break.
Connection should be unconditionally threadsafe as defined in Effective Java by Joshua Bloch Item70: Document thread safety. Also see the commit message

PR #186, opened 2017-11-22
Unfinalize classes to allow to mocking/subclassing

### LINKED ISSUE #185: final Classes prevent mocking or Subclassing
I'm trying to integrate tchannel-java into an application, injecting the `SubChannel` into a Class to interact with another tchannel service.

But, there's effectively no way for me to test this class - I can't mock final classes, or make my own in-memory implementation of the `SubChannel` class.

I tried creating a non-final Wrapper for `SubChannel`, only exposing the relevant methods I intend to use, and a delegating to the actual `SubChannel`, but that's a rabbit hole since `TFuture`, `ThriftResponse` are all final.

I propose making all these class non-final - it would be a backwards compatible change, allow end users to integrate with the libraries with tests for their logic, and let users customize the implementations as needed.

achals:
Fixes #185.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java
@@ -58,7 +58,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Random;
 
-public final class SubChannel {
+public class SubChannel {
 
     private final String service;
     private final @NotNull TChannel topChannel;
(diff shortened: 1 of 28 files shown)


alshopov:
I get your point about mocking the classes. Since I am doing most of the things `final` - my reason for that is to explicitly show users that the classes are not meant for extension rather than hamper people. PowerMock really allows you to mock even final classes and static members.
If @yborovikov is fine with your change I will not block it.
I believe most proper way will be to extract interfaces, make these final classes implement them to allow for both `final` and `mockito`-ability. I am not sure whether this will be worth it.

achals:
@yborovikov Bump, could you take a look at this when you had the chance?

PR #192, opened 2018-01-21
[tchannel-java] Consolidate to gson 2.8.2 dependency

dray92:


pom.xml
```diff
@@ -164,12 +164,6 @@
                 <groupId>com.uber.jaeger</groupId>
                 <artifactId>jaeger-core</artifactId>
                 <version>0.23.0</version>
-                <exclusions>
```
yurishkuro: why not keep this exclusion? Since tchannel directly depends on gson there's always going to be a contention between what tchannel wants and what jaeger brings in transitively.

dray92: I don't have an opinion either way - I am going to bring back the exclusion cause your argument makes sense to me

yborovikov: today tchannel depends on gson, tomorrow it doesn't (and something somewhere breaks because we excluded this something's dependency). unless a dependency is deeply troubled one, excluding its dependencies should not be done casually. for shared dependency version management there is `dependencyManagement` section.

PR #196, opened 2018-02-14
Add OpenTracingContext implementation

yborovikov:
This PR adds an `io.opentracing.ScopeManager`-based implementation of `TracingContext`.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -33,6 +33,7 @@ import com.uber.tchannel.handlers.MessageFragmenter;
 import com.uber.tchannel.handlers.RequestRouter;
 import com.uber.tchannel.handlers.ResponseRouter;
 import com.uber.tchannel.messages.Request;
+import com.uber.tchannel.tracing.OpenTracingContext;
 import com.uber.tchannel.tracing.TracingContext;
 import com.uber.tchannel.utils.TChannelUtilities;
 import io.netty.bootstrap.Bootstrap;
@@ -109,7 +110,9 @@ public final class TChannel {
         this.timer = builder.timer;
         this.clientMaxPendingRequests = builder.clientMaxPendingRequests;
         this.tracer = builder.tracer;
-        this.tracingContext = builder.tracingContext;
+        this.tracingContext = builder.tracingContext == null
+            ? tracer == null ? new TracingContext.Default() : new OpenTracingContext(tracer.scopeManager())
+            : builder.tracingContext;
     }
 
     public String getListeningHost() {
@@ -247,7 +250,7 @@ public final class TChannel {
         private static final int WRITE_BUFFER_HIGH_WATER_MARK = 32 * 1024;
 
         private Tracer tracer;
-        private TracingContext tracingContext = new TracingContext.Default();
+        private TracingContext tracingContext;
         private ExecutorService executorService = null;
 
         public Builder(@NotNull String service) {
--- a/tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2015 Uber Technologies, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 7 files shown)


tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
```diff
@@ -109,7 +110,9 @@ private TChannel(@NotNull Builder builder) {
         this.timer = builder.timer;
         this.clientMaxPendingRequests = builder.clientMaxPendingRequests;
         this.tracer = builder.tracer;
-        this.tracingContext = builder.tracingContext;
+        this.tracingContext = builder.tracingContext == null
```
yborovikov: `tracingContext` now defaults to `OpenTracingContext` backed by native tracer `scopeManager` (instead of useless no-op implementation).

tchannel-core/src/test/java/com/uber/tchannel/tracing/AsyncTracingContextTest.java
```diff
@@ -1,132 +1,40 @@
+/*
```
yborovikov: main tests functionality was extracted to shared `AsyncTracingContextTestBase` parent class.

tchannel-core/src/test/java/com/uber/tchannel/tracing/AsyncTracingContextTestBase.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.sameInstance;
+
+/**
+ * Base for tests of asynchronous request callback tracing context propagation for various implementations of
+ * {@link TracingContext} interface.
+ *
+ * @author yegor 2017-11-10.
+ */
+public abstract class AsyncTracingContextTestBase {
```
yborovikov: tests code moved here from `AsyncTracingContextTest`.

tchannel-core/src/test/java/com/uber/tchannel/tracing/TracingContextTest.java
```diff
@@ -22,81 +22,16 @@
...
-        Span span = tracer.buildSpan("test").startManual();
-        tracingContext.pushSpan(span);
-        assertTrue(tracingContext.hasSpan());
-        assertEquals(span, tracingContext.currentSpan());
-        assertEquals(span, tracingContext.popSpan());
-        assertFalse(tracingContext.hasSpan());
+/**
+ * Tests {@link TracingContext.ThreadLocal} implementation of {@link TracingContext} interface.
+ */
+public class TracingContextTest extends TracingContextTestBase {
```
yborovikov: main tests functionality was extracted into the shared `TracingContextTestBase` parent class.

tchannel-core/src/test/java/com/uber/tchannel/tracing/TracingContextTestBase.java
```diff
@@ -0,0 +1,97 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+/**
+ * Base for various {@link TracingContext} implementations tests.
+ *
+ * @author yegor 2018-02-13.
+ */
+public abstract class TracingContextTestBase {
```
yborovikov: tests code moved here from `TracingContextTest` - for sharing and unification.

tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
```diff
@@ -0,0 +1,94 @@
...
+        if (scope == null) {
+            throw new EmptyStackException();
+        } else {
+            return scope.span();
+        }
+    }
+
+    @Override
+    public @NotNull Span popSpan() throws EmptyStackException {
+        try (Scope scope = scopeManager.active()) {
```
yurishkuro: this will try to close the scope, right? what happens if the scope is null?

yborovikov: (1) yes
(2) nothing (per JLS)

yurishkuro:
My main concern is with TracingContext.popSpan() being implemented via scope.close(). The latter is not equivalent to popping the span, because depending on how the scope was created it may also finish the span on close.

One option is to remove TracingContext altogether and just use tracer.scopeManager() as intended. But that would be a breaking change (although pulling opentracing 0.31 was already a breaking change).

Another option is to continue accepting TracingContext as optional (and deprecated) and wrap it into an adapter implementing ScopeManager, so that the rest of tchannel code is just using the scope manager.

yborovikov:
I agree - `TracingContext` has outlived its welcome.

As for side-effects of `Scope.close()` - these probably should be fixed in opentracing itself. By either removing the behavior (not sure what the rationale was, to begin with - maybe you know the context) or by introducing some sort of a counter that would track activation/deactivation events (which is not likely to work properly even in ideal conditions).

Good thing is, though, that tchannel _only_ pops spans that it _itself_ pushed (unless. of course, the context got corrupted but this is a totally different and much sadder story) - and `OpenTracingContxt.push()` implementation doesn't set the `finishSpanOnClose` flag. So we're safe from that perspective.

yurishkuro:
> Good thing is, though, that tchannel only pops spans that it itself pushed

ah, that's true. In this case we should be fine.

> As for side-effects of Scope.close() - these probably should be fixed in opentracing itself. 

That behavior is by design.  In fact, the large change from ActiveSpanManager to ScopeManager in 0.31 was specifically done to remove reference counting that was implemented in 0.30, as it caused more unexpected behavior. 

:shipit:

PR #198, opened 2019-05-26
upgrade java compiler source to 1.8 and add a type annotations for Re…

henrywoo:
To justify the use of `response.close()` which people often use, we need to specify `NotNull` for Response, so I added a type annotation. It failed with an error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:compile (default-compile) on project tchannel-core: Compilation failure
[ERROR] /home/henry/share/company/uber/tchannel-java/tchannel-core/src/main/java/com/uber/tchannel/api/SubChannel.java:[245,27] type annotations are not supported in -source 1.7
[ERROR]   (use -source 8 or higher to enable type annotations)
[ERROR] 
```

This conflicts with the comment in pom.xml:
```
<!-- store method parameter names in the bytecode (java 8+) -->
```

It seems we forgot to specify java 8. So I modified it. That is how this diff comes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,7 +83,7 @@
         <!-- store method parameter names in the bytecode (java 8+) -->
         <maven.compiler.parameters>true</maven.compiler.parameters>
         <maven.compiler.showWarnings>true</maven.compiler.showWarnings>
-        <maven.compiler.source>1.7</maven.compiler.source>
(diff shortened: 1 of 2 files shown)


henrywoo:
The Travis CI build failed because it is still using Java 1.7. Could anybody in charge of this project please upgrade it to 1.8? Thanks.

PR #199, opened 2019-07-09
Upgrade opentracing & jaeger lib

guo0693:
- upgrade opentracing to 0.32.0
- upgrade jaeger to 1.1.1
- fix the infinite loop bug when closing tracing context due to the NoopScopeManager always return non-null instance
Base on the comment: https://code.uberinternal.com/D2999387?id=9423683#inline-21546995

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,6 +93,7 @@
         <jmh.version>1.19</jmh.version>
         <se.eris.notnull.instrument>false</se.eris.notnull.instrument><!-- off by default -->
         <slf4j.version>1.7.25</slf4j.version>
+        <opentracing.version>0.32.0</opentracing.version>
 
     </properties>
 
@@ -163,12 +164,17 @@
             <dependency>
                 <groupId>io.opentracing</groupId>
                 <artifactId>opentracing-api</artifactId>
-                <version>0.31.0</version>
+                <version>${opentracing.version}</version>
             </dependency>
             <dependency>
-                <groupId>com.uber.jaeger</groupId>
+                <groupId>io.opentracing</groupId>
+                <artifactId>opentracing-noop</artifactId>
+                <version>${opentracing.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.jaegertracing</groupId>
                 <artifactId>jaeger-core</artifactId>
-                <version>0.23.0</version>
+                <version>0.35.5</version>
             </dependency>
             <dependency>
                 <groupId>org.jetbrains</groupId>
--- a/tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
@@ -52,42 +52,46 @@ public class OpenTracingContext implements TracingContext {
     @Override
     @SuppressWarnings("resource")
     public void pushSpan(@NotNull Span span) {
-        scopeManager.activate(span, false);
+        scopeManager.activate(span);
     }
 
     @Override
     @SuppressWarnings("resource")
     public boolean hasSpan() {
-        return scopeManager.active() != null;
+        return scopeManager.activeSpan() != null;
     }
 
     @Override
     public @NotNull Span currentSpan() throws EmptyStackException {
-        @SuppressWarnings("resource") Scope scope = scopeManager.active();
-        if (scope == null) {
+        Span span = scopeManager.activeSpan();
+        if (span == null) {
             throw new EmptyStackException();
-        } else {
-            return scope.span();
         }
+        return span;
     }
 
     @Override
     public @NotNull Span popSpan() throws EmptyStackException {
+        //auto close scope acts as if we pop the span
         try (Scope scope = scopeManager.active()) {
             if (scope == null) {
                 throw new EmptyStackException();
-            } else {
-                return scope.span();
             }
+            return scope.span();
         }
     }
 
     @Override
     @SuppressWarnings("resource")
     public void clear() {
-        Scope scope;
-        while ((scope = scopeManager.active()) != null) {
+        Scope lastScope = null;
+        Scope scope = scopeManager.active();
+        //NoopScopeManager will always return the same INSTANCE as active scope
+        //to avoid infinite loop we need to check if we are closing the same instance again and again
+        while (scope != null && lastScope != scope) {
             scope.close();
+            lastScope = scope;
+            scope = scopeManager.active();
         }
     }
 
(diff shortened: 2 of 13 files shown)


guo0693:
tchannel seems to be using google maven central, java monorepo is not publishing to that artifactory, going to figure this out.

pom.xml
```diff
@@ -163,12 +164,17 @@
...
+            <dependency>
+                <groupId>io.opentracing</groupId>
+                <artifactId>opentracing-noop</artifactId>
+                <version>${opentracing.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.uber.jaeger</groupId>
                 <artifactId>jaeger-core</artifactId>
-                <version>0.23.0</version>
+                <version>1.1.1-1136-e7ae7fd51c4</version>
```
vprithvi: We should migrate this to `io.jaegertracing:jaeger-core:0.35.5`, which is the latest OSS Jaeger version

yurishkuro: tchannel shouldn't even have this dependency, only opentracing

tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
```diff
@@ -85,6 +86,10 @@ public boolean hasSpan() {
     @Override
     @SuppressWarnings("resource")
     public void clear() {
+        if(scopeManager instanceof NoopScopeManager) {
```
yurishkuro: Why is this check needed? Please add comment.

guo0693: Code updated with a better check logic so that if other type of ScopeManager acts in the same way as NoopScopeManager, the logic can also avoid infinite loop.

Comment added.

tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
```diff
@@ -52,31 +52,31 @@ public OpenTracingContext(@NotNull ScopeManager scopeManager) {
...
     public @NotNull Span popSpan() throws EmptyStackException {
-        try (Scope scope = scopeManager.active()) {
-            if (scope == null) {
-                throw new EmptyStackException();
-            } else {
+        if (!hasSpan()) {
+            throw new EmptyStackException();
+        } else {
+            //auto close scope acts as if we pop the span
+            try (Scope scope = scopeManager.active()) {
```
guo0693: @yurishkuro @vprithvi I noticed that opentracing has marked scopeManager.active() as deprecated, they provide activeSpan() to access the span. However here we need the access to scope so that when the current span is popped, the scope can be closed and thus threadLocal can set the saved scope as active. I am afraid if opentracing remove the act() method, it will break this OpentracingContext, that means we may also need to deprecate this custom context implemented provided by tchannel and ask tchannel users to use scope manager instead, which may be another big task.

tchannel-crossdock/src/test/java/com/uber/tchannel/crossdock/TraceBehaviorTest.java
```diff
@@ -124,8 +125,8 @@ public void TestTraceBehavior() throws Exception {
         logger.info("Response: {}", response);
         span.finish();
 
-        SpanContext spanContext = (SpanContext) span.context();
-        String traceId = String.format("%x", spanContext.getTraceId());
+        SpanContext spanContext = span.context();
+        String traceId = String.format("%s", spanContext.toTraceId());
```
yurishkuro: `toTraceId` already returns a string, on need to call format()

pom.xml
```diff
@@ -163,12 +164,17 @@
...
-                <groupId>com.uber.jaeger</groupId>
+                <groupId>io.opentracing</groupId>
+                <artifactId>opentracing-noop</artifactId>
+                <version>${opentracing.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.jaegertracing</groupId>
                 <artifactId>jaeger-core</artifactId>
-                <version>0.23.0</version>
+                <version>0.35.5</version>
```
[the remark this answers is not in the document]
guo0693: tchannel-core pom is updated such that jaeger-core is just test time dependency

tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
yurishkuro: OpenTracing API my be  frozen soon in favor  of OpenTelemetry. But yet, the right thing would be to change tchannel implementation to use scopes. I just don't think it's pressing.

tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
```diff
@@ -52,42 +52,46 @@ public OpenTracingContext(@NotNull ScopeManager scopeManager) {
...
         }
+        return span;
     }
 
     @Override
     public @NotNull Span popSpan() throws EmptyStackException {
+        //auto close scope acts as if we pop the span
         try (Scope scope = scopeManager.active()) {
             if (scope == null) {
                 throw new EmptyStackException();
```
yurishkuro: what is going to happen to `try` with null?

tchannel-core/src/test/java/com/uber/tchannel/tracing/TracingPropagationTest.java
```diff
@@ -323,9 +320,9 @@ private static TraceResponse callDownstreamThrift(String remainingEncodings, Str
 
     @Test
     public void testPropagation() throws Exception {
-        Span span = (Span) tracer.buildSpan("root").startManual();
-        SpanContext context = span.context();
-        String traceId = String.format("%x", context.getTraceId());
+        Span span = tracer.buildSpan("root").start();
+        JaegerSpanContext context = (JaegerSpanContext) span.context();
+        String traceId = context.getTraceId();
```
yurishkuro: I suggest using `toTraceId()` consistently everywhere, not mixing get- and to-

tchannel-core/src/main/java/com/uber/tchannel/tracing/OpenTracingContext.java
guo0693: A resource is closed only if it initialized to a non-null value. When scope is null, it won't be closed and there won't be NPE after the EmptyStackException is thrown. There is unit test that covers this scenario.

tchannel-core/src/test/java/com/uber/tchannel/tracing/TracingPropagationTest.java
guo0693: For this line, we can use SpanContext.toTraceId(), but there are a few cases where JaegerSpanContext is needed e.g. the code needs to know if the span is sampled or not (Line:231).

Do you want to have this?
`
       	 response.traceId = span.context().getTraceId();
         JaegerSpanContext context = (JaegerSpanContext) span.context();
          response.sampled = context.isSampled();`

guo0693: Make sense. I thought SpanContext provides toTraceId while JaegerSpanContext provides getTraceId. Just found out both have toTraceId function.

Updated.

PR #201, opened 2019-11-15
Allow customizing child group thread pool size

Garciat:
The child group initialization code calls the default constructors of EpollEventLoopGroup and NioEventLoopGroup, which will construct ForkJoinPools with a maximum # of threads determined by `Runtime.getRuntime().availableProcessors()`.

For containerized applications in older JDK versions, the JVM is unable to correctly determine the number of available CPU cores. So these FJPs will likely by oversized in these scenarios.

This change allows the user to specify the number of threads used for the boss & child group thread pools. (This is already possible with `setBossGroup`/`setChildGroup` but is more convoluted.)

Extra change: also names the corresponding threads for debugging purposes.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/api/TChannel.java
@@ -21,6 +21,7 @@
  */
 package com.uber.tchannel.api;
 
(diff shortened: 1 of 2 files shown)


chornyi:
lgtm

PR #210, opened 2019-12-14
Make ThriftRequest.Builder#build() and ThriftResponse.Builder#build() auto release ByteBuf if validation fails

denyska:
Both `ThriftRequest.Builder#validate()` and `ThriftResponse.Builder#validate()` are stateful, i.e. they modify state of the Builder object.  

`ThriftRequest.Builder#validate()`:

```
        @Override
        public Builder<T> validate() {
            super.validate();
            this.validateHeader();
            this.validateBody();
            return this;
        }

        private Builder<T> validateHeader() {
            if (arg2 != null) {
                headers = null;
                return this;
            }

            arg2 = serializer.encodeHeaders(this.headers, argScheme);

            return this;
        }

        private Builder<T> validateBody() {
            if (arg3 != null) {
                body = null;
                return this;
            }

            if (body == null) {
                arg3 = TChannelUtilities.emptyByteBuf;
            } else {
                arg3 = serializer.encodeBody(this.body, argScheme);
            }

            return this;
        }
```
Hence if body validation fails we end up with native memory leak in ByteBuf: ByteBuf is already allocated for the headers.

**Builders** can be re-used once the problem with headers or body fixed. Added tests for it.

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/messages/ThriftRequest.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/messages/ThriftRequest.java
@@ -79,8 +79,6 @@ public class ThriftRequest<T> extends EncodedRequest<T> {
          *
          * Args above will be auto-released if validation fails.
          *
-         * Note: Don't call it again if fails.
-         * <br/>unless header/body are re-populated this method will fail if called after the initial failure.
          */
         public ThriftRequest<T> build() {
             ThriftRequest<T> result;
--- a/tchannel-core/src/main/java/com/uber/tchannel/messages/ThriftRequest.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/messages/ThriftRequest.java
@@ -58,14 +58,43 @@ public class ThriftRequest<T> extends EncodedRequest<T> {
             this.argScheme = ArgScheme.THRIFT;
         }
 
+        /**
+         * Validates payload and populates {@link #arg1}, {@link #arg2}, {@link #arg3}.
+         *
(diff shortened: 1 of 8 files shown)


tchannel-core/src/main/java/com/uber/tchannel/messages/ThriftRequest.java
```diff
@@ -58,14 +58,43 @@ public Builder(String service, ByteBuf arg1) {
...
+         * Validates payload, populates {@link #arg1}, {@link #arg2}, {@link #arg3} and builds {@link ThriftRequest}.
+         *
+         * Args above will be auto-released if validation fails.
+         *
+         * Note: Don't call it again if fails.
+         * <br/>unless header/body are re-populated this method will fail if called after the initial failure.
+         */
         public ThriftRequest<T> build() {
-            return new ThriftRequest<>(this.validate());
+            ThriftRequest<T> result;
```
raviagarwal7: could we create a copy of the builder and call build on that? this will allow reusing the builder even if validation fails.

Garciat:
If we switch the `build()` method from calling the request constructor with `this` to calling the constructor with the exploded argument list, we can perform the encoding and handle the exceptions on each `build()` call, without any side-effects.

What do you think?

tchannel-core/src/main/java/com/uber/tchannel/messages/ThriftRequest.java
denyska: I made Request.validateHeader and Request.validateBody idempotent: they can be called number of times w/o producing side effect if arg2/arg3 already constructed.

PR #220, opened 2020-03-13
Inject/extract trace information into/from request

ChuntaoLu:
This PR does the following:
- Inject tracing info into outbound request (call request frame) besides header;
- Extract tracing info from inbound request when tracing info isn't present in headers.

Rationale:
-  TChannel has built-in support for propagating [tracing info](https://github.com/uber/tchannel/blob/master/docs/protocol.md#tracing) with reserved bytes in [call req frame](https://github.com/uber/tchannel/blob/master/docs/protocol.md#call-req-type-0x03).
- TChannel-go [creates](https://github.com/uber/tchannel-go/blob/dev/inbound.go#L86) spans with tracing info in call req frame (not from headers), so if a tchannel-java client calls tchannel-go server with trace info carried in headers, the server span would be missing since request form tchannel-java client does not contain tracing info. [Here](https://code.uberinternal.com/T5310765) is an Uber internal reference.

### DIFF
--- a/tchannel-core/pom.xml
+++ b/tchannel-core/pom.xml
@@ -75,6 +75,10 @@
             <artifactId>guava</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.jaegertracing</groupId>
+            <artifactId>jaeger-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.opentracing</groupId>
             <artifactId>opentracing-api</artifactId>
         </dependency>
@@ -103,10 +107,6 @@
             <groupId>io.netty</groupId>
             <artifactId>netty-codec-http</artifactId>
         </dependency>
-        <dependency>
-            <groupId>org.apache.thrift</groupId>
-            <artifactId>libthrift</artifactId>
-        </dependency>
 
         <!-- test dependencies -->
         <dependency>
@@ -117,11 +117,6 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>io.jaegertracing</groupId>
-            <artifactId>jaeger-core</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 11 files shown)


tchannel-core/pom.xml
```diff
@@ -103,10 +107,6 @@
             <groupId>io.netty</groupId>
             <artifactId>netty-codec-http</artifactId>
         </dependency>
-        <dependency>
```
ChuntaoLu: FYI, this is an duplicate dependency.

tchannel-core/src/main/java/com/uber/tchannel/tracing/Tracing.java
```diff
@@ -89,7 +91,18 @@ private Tracing() {}
...
+                jaegerSpanContext.getParentId(),
+                // tchannel only support 64bit IDs, https://github.com/uber/tchannel/blob/master/docs/protocol.md#tracing
+                jaegerSpanContext.getTraceIdLow(),
+                jaegerSpanContext.getFlags());
+            request.setTrace(trace);
+        }
+
         // if request has headers, inject tracing context
         if (request instanceof TraceableRequest) {
             TraceableRequest traceableRequest = (TraceableRequest) request;
```
denyska: Should we also try to extract trace from the headers and inject into trace field?

ChuntaoLu: This is actually the outbound path, where the headers don't have the trace info yet. The existing code in this method creates a child span and injects it into headers, what I added here is to also inject the same trace info into request trace fields.

chornyi:
This breaks if `TChannel` is configured with a `MockTracer` instance:

```
java.lang.ClassCastException: io.jaegertracing.internal.JaegerSpanContext cannot be cast to io.opentracing.mock.MockSpan$MockContext
	at io.opentracing.mock.MockTracer$SpanBuilder.addReference(MockTracer.java:340)
	at io.opentracing.mock.MockTracer$SpanBuilder.asChildOf(MockTracer.java:320)
	at io.opentracing.mock.MockTracer$SpanBuilder.asChildOf(MockTracer.java:307)
	at com.uber.tchannel.tracing.Tracing.startInboundSpan(Tracing.java:172)
	at com.uber.tchannel.handlers.RequestRouter.sendRequestToAsyncHandler(RequestRouter.java:248)
	at com.uber.tchannel.handlers.RequestRouter.channelRead0(RequestRouter.java:134)
	at com.uber.tchannel.handlers.RequestRouter.channelRead0(RequestRouter.java:57)
```

I wonder if this should be implemented similarly to `tracer.extract(Format.Builtin.TEXT_MAP, carrier);` so that the `Tracer` can instantiate the right `SpanContext` subtype.

PR #223, opened 2020-04-02
Tag span with error = true in case of ResponseCode.Error

chornyi:
This would help with Thrift exceptions which are returned in the body and treated as successes.

I could not write a test due to https://github.com/uber/tchannel-java/pull/220#issuecomment-607660999

### DIFF
--- a/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
+++ b/tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
@@ -28,6 +28,7 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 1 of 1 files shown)


tchannel-core/src/main/java/com/uber/tchannel/handlers/RequestRouter.java
```diff
@@ -251,13 +254,18 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
...
+
                 closeRequestAndSpan();
             }
 
             @Override
             public void onFailure(@NotNull Throwable e) {
                 if (span != null) {
-                    span.log(ImmutableMap.of("exception", e));
+                    span.setTag(Tags.ERROR, true);
+                    span.log(ImmutableMap.of(Fields.ERROR_OBJECT, e));
```
chornyi, line 268: Based on https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerSpan.java#L285-L309

denyska:
LGTM

PR #225, opened 2020-07-30
Fix for continuous timeouts

ankuriitg:
The timeouts counter is used to track continuous timeouts and if it reaches above the resetOnTimeoutLimit, then it resets the connection. There is a bug in the timeouts counter as it is always incremented on every timeout because of which it is not tracking continuous timeouts but cumulative timeouts since the start of instance.

This fix resets the counter if we receive any response other than timeout, thus fixing the bug.

Testing Done:
Added a junit test to ensure the logic works now


sgolecha:
@ankuriitg  - does this mean this will reduce the alert noise?

ankuriitg:
> @ankuriitg - does this mean this will reduce the alert noise?

No, this has nothing to do with our service alerts.

